PR #2, opened 2026-03-04
Add mongodb-query skill with testing infrastructure MCP-425

betsybutton:
Introduces a new skill for generating MongoDB queries and aggregation pipelines from natural language descriptions. Includes comprehensive testing workspace with fixtures, evaluation cases, and performance benchmarks.

  ## Skill Information                                                      
                  
  **Skill Name:** mongodb-natural-language-querying                                        
  **Skill Directory:** `skills/mongodb-natural-language-querying`                               
                                                                            
  ## Use Case

  This skill addresses the challenge of generating MongoDB queries and
  aggregation pipelines from natural language descriptions. MongoDB users
  often need to:
  - Translate business requirements into correct MongoDB query syntax
  - Convert SQL-like thinking into MongoDB's document-oriented query
  language
  - Build complex aggregation pipelines with multiple stages
  - Optimize queries based on available indexes and collection schemas
  - Handle geospatial queries, date-relative filtering, and multi-collection
   joins

  The skill automates this translation by leveraging collection schemas,
  sample documents, and index information to generate accurate, optimized
  queries.

  ## Value Proposition

  This skill is valuable for MongoDB users because it:
  - **Reduces learning curve**: Users don't need to memorize MongoDB query
  syntax or aggregation pipeline stage ordering
  - **Context-aware generation**: Uses actual collection schemas and sample
  data to generate queries that match the real data structure
  - **Optimization guidance**: Considers available indexes when suggesting
  query patterns
  - **Handles complexity**: Supports advanced operations like geospatial
  queries, text search, regex operations, and $lookup joins
  - **Multi-language support**: Works with queries in multiple languages
  (validated with Spanish)
  - **Error prevention**: Validates field names and types against schemas
  before generating queries

  Tested across 8 diverse scenarios with a 93.75% success rate.

  ## Special Considerations

  **Prerequisites:**
  - MongoDB MCP server must be configured with a valid connection string
  - Requires `.mcp.json` configuration with `mongodb-mcp-server` entry
  - Collection must exist and be accessible for schema/sample data retrieval

  **MCP Server Configuration Example:**
  ```json
  {
    "mcpServers": {
      "mongodb": {
        "command": "npx",
        "args": ["-y", "mongodb-mcp-server@latest"],
        "env": {
          "MDB_MCP_CONNECTION_STRING": "your-connection-string-here"
        }
      }
    }
  }  
```

## Limitations:
  - Does not support MongoDB Atlas Search ($search) operator - will suggest
  text indexes instead
  - Requires database and collection names to be provided or discoverable

  ## Validation Prompts

  1. Show me all movies from the year 2000 in the netflix.movies collection,
   but only return the title and year fields, sorted by title
  2. Find cocktail bars in berlin.cocktailbars within 1000 meters of
  coordinates [13.4050, 52.5200]
  3. Calculate the mode (most common value) of the year field in
  netflix.movies collection
  4. Encuentra todas las películas de la década de 1990 en la colección
  netflix.movies (Spanish: Find all movies from the 1990s)
  5. Join netflix.movies with netflix.comments to show movie titles along
  with their comment counts and top 3 most recent comments

 ##  Author Self-Validation

  - Verified structure and format requirements
  - Tested skill with LLMs using the five prompts above
  - Checked content quality and novelty
  - Validated all links and file references
  - (Optional but recommended) Ran skill-validator locally

## SME Review

**SME:** @asya999

## Additional Context
  Testing Infrastructure:
  - Includes comprehensive test workspace at
 mongodb-natural-language-querying/mongodb-query-workspace/
  - 5 fixture collections with 45 sample documents total
  - 8 evaluation test cases covering diverse query types
  - Benchmark results showing 93.75% success rate (7.5/8 tests passing)
  - Load fixture scripts for easy test environment setup
  - Documented test results and performance analysis

  Notable Test Scenarios:
  - Simple find queries with filters
  - Complex queries with projections, sorting, and limits
  - Geospatial queries with $near operator
  - Aggregation pipelines with regex text processing
  - Multi-collection joins using $lookup
  - Date-relative queries (e.g., "last year")
  - Non-English prompt handling

The one partially passing test (text search) is a known limitation where the skill suggests text indexes instead of $search operator, which is acceptable behavior.

### CODE REVIEW

File: skills/mongodb-query-generator/SKILL.md [hunk 1]
```diff
@@ -0,0 +1,239 @@
...
+
+### 2. Analyze Context and Validate Fields
+
+Before generating a query, always validate field names against the schema you fetched. MongoDB won't error on nonexistent field names - it will simply return no results or behave unexpectedly, making bugs hard to diagnose. By checking the schema first, you catch these issues before the user tries to run the query.
+
+Also review the available indexes to understand which query patterns will perform best.
+
+### 3. Choose Query Type: Find vs Aggregation
+
+Prefer find queries over aggregation pipelines because find queries are simpler, faster, and easier for other developers to understand. Find queries also have better performance characteristics for simple filtering and sorting since they avoid the aggregation framework overhead.
```
asya999: find is not faster than equivalent agg

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+For aggregation pipelines:
+- ✅ `"[{ $match: { status: 'active' } }]"` (string)
+- ❌ `[{ $match: { status: 'active' } }]` (array)
+
+## Best Practices
+
+### Query Quality
+1. **Use indexes efficiently** - Structure filters to leverage available indexes:
+   - Check collection indexes before generating the query
+   - Order filter fields to match index key order when possible
```
asya999: that's not a thing - order of filter fields does not matter.

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+   - Use equality matches before range queries (matches index prefix behavior)
```
asya999: in the query it does not matter

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+   - Avoid operators that prevent index usage: `$where`, `$text` without text index, `$ne`, `$nin` (use sparingly)
```
asya999: $where should never be used, $expr should be used only when necessary, I don't see a problem with inequality

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+   - For compound indexes, use leftmost prefix when possible
+   - If no relevant index exists, mention this in your response (user may want to create one)
+2. **Project only needed fields** - Reduce data transfer with projections
+3. **Validate field names** against the schema before using them
+4. **Handle edge cases** - Consider null values, missing fields, type mismatches
+5. **Use appropriate operators** - Choose the right MongoDB operator for the task:
+   - `$eq`, `$ne`, `$gt`, `$gte`, `$lt`, `$lte` for comparisons
+   - `$in`, `$nin` for membership tests
```
asya999: that's not what `$in` `$nin` do.  They are just equality operators with multiple possible values.

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+   - `$and`, `$or`, `$not`, `$nor` for logical operations
+   - `$regex` for text pattern matching
```
asya999: where should we mention left anchored being preferred?

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+2. **Project only needed fields** - Reduce data transfer with projections
+3. **Validate field names** against the schema before using them
+4. **Handle edge cases** - Consider null values, missing fields, type mismatches
+5. **Use appropriate operators** - Choose the right MongoDB operator for the task:
+   - `$eq`, `$ne`, `$gt`, `$gte`, `$lt`, `$lte` for comparisons
+   - `$in`, `$nin` for membership tests
+   - `$and`, `$or`, `$not`, `$nor` for logical operations
+   - `$regex` for text pattern matching
+   - `$exists` for field existence checks
+   - `$type` for type validation
```
asya999: I'm a bit concerned about these - can we at least make sure they are never added when they are redundant of other conditions already in the query?

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+   - `$exists` for field existence checks
+   - `$type` for type validation
+
+### Aggregation Pipeline Quality
+1. **Filter early** - Use `$match` as early as possible to reduce documents
+2. **Project early** - Use `$project` to reduce field set before expensive operations
```
asya999: That's an anti-pattern.  $project should only be used at the end to correctly shape returned documents to client

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Aggregation Pipeline Quality
+1. **Filter early** - Use `$match` as early as possible to reduce documents
+2. **Project early** - Use `$project` to reduce field set before expensive operations
+3. **Limit when possible** - Add `$limit` after `$sort` when appropriate
+4. **Use indexes** - Ensure `$match` and `$sort` stages can use indexes:
+   - Place `$match` stages at the beginning of the pipeline
+   - Initial `$match` and `$sort` stages can use indexes if they precede any stage that modifies documents
```
asya999, line 140: that's not strictly true... but there's not a simple description that's possible here, probably better leave it open ended.

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+3. **Limit when possible** - Add `$limit` after `$sort` when appropriate
+4. **Use indexes** - Ensure `$match` and `$sort` stages can use indexes:
+   - Place `$match` stages at the beginning of the pipeline
+   - Initial `$match` and `$sort` stages can use indexes if they precede any stage that modifies documents
+   - Structure `$match` filters to align with available indexes
+   - Avoid `$project`, `$unwind`, or other transformations before `$match` when possible
```
asya999: $project should be at the end.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+   - Structure `$match` filters to align with available indexes
+   - Avoid `$project`, `$unwind`, or other transformations before `$match` when possible
+5. **Optimize `$lookup`** - Consider denormalization for frequently joined data
```
asya999, line 143: ensure indexes on fields in foreign collection being joined

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+- Identify field naming conventions (camelCase, snake_case, etc.)
+- Detect common patterns (e.g., status enums, category values)
+- Estimate cardinality for grouping operations
+- Validate that your query will work with real data
+
+## Common Pitfalls to Avoid
+
+1. **Using nonexistent field names** - Always validate against schema first. MongoDB won't error; it just returns no results.
+2. **Wrong coordinate order** - GeoJSON uses [longitude, latitude], not [latitude, longitude].
+3. **Choosing aggregation when find suffices** - Aggregation adds overhead; use find for simple queries.
```
asya999: this is not the case.

skills/mongodb-query-generator/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+4. **Missing index awareness** - Structure queries to leverage indexes. If no index exists for key filters, mention this to the user.
+5. **Type mismatches** - Check schema to ensure operators match field types (e.g., don't use `$gt` on strings when comparing alphabetically).
```
asya999: wait, why not? that's how it works with strings...

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+3. Generate query:
+```json
+{
+  "query": {
+    "filter": "{ status: 'active', age: { $gt: 25 } }",
+    "sort": "{ registrationDate: -1 }"
+  }
+}
+```
+
```
asya999, line 198: should there be two more steps here - suggest index if there isn't one and suggest adding limit if a collection is relatively large?

asya999 (commented):
left some comments

skills/mongodb-query-generator/SKILL.md
betsybutton: I think we can mention it here - good point

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Okay, I won't make an edit for now - lmk if you have a better idea of how to handle this case

testing/mongodb-query-generator/mongodb-query-workspace/load-fixtures.js
```diff
@@ -0,0 +1,91 @@
...
+    .replace(/\$oid:\s*'([^']+)'/g, '"$1"') // Convert $oid format
+    .replace(/ObjectId\(/g, '')
+    .replace(/\)/g, '')
+    .replace(/,\s*\]/g, ']') // Remove trailing commas
+    .replace(/,\s*\}/g, '}');
+
+  try {
+    const docs = eval('(' + jsonStr + ')');
+    // Convert _id fields with $oid to ObjectId
+    return docs.map(doc => {
```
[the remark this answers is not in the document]
betsybutton: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/mongodb/agent-skills/pull/2#discussion_r2919126974)

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+**For Aggregation Pipelines**, generate an array of stage objects.
+
+**Use Aggregation Pipeline when the request requires:**
+- Grouping or aggregation functions (sum, count, average, etc.)
+- Multiple transformation stages
+- Computed fields or data reshaping
+- Joins with other collections ($lookup)
+- Array unwinding or complex array operations
+- Text search with scoring
+
```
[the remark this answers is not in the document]
betsybutton: Need to reconcile this with the search skill

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+- `sort` - Sort specification (optional)
+- `skip` - Number of documents to skip (optional)
+- `limit` - Number of documents to return (optional)
+- `collation` - Collation specification (optional)
+
+**Use Find Query when:**
+- Simple filtering on one or more fields
+- Basic sorting and limiting
+- Field projection only
+- No data transformation needed
```
asya999: find can do field transformation same as agg... (projection language is the same in both). This probably means to say ... no document tranformations?  Actually what _is_ it trying to say?

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+For aggregation pipelines:
+- ✅ `"[{ $match: { status: 'active' } }]"` (string)
+- ❌ `[{ $match: { status: 'active' } }]` (array)
+
+## Best Practices
+
+### Query Quality
+1. **Use indexes efficiently** - Structure filters to leverage available indexes:
+   - Check collection indexes before generating the query
+   - Avoid operators that prevent index usage: `$where`
```
asya999: I think we need to say "never use $where".
Avoid should be reserved for `$expr` when it's not necessary.  $where is NEVER necessary.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+- ❌ `{ age: { $gte: 25 } }` (object)
+
+For aggregation pipelines:
+- ✅ `"[{ $match: { status: 'active' } }]"` (string)
+- ❌ `[{ $match: { status: 'active' } }]` (array)
+
+## Best Practices
+
+### Query Quality
+1. **Use indexes efficiently** - Structure filters to leverage available indexes:
```
asya999: ok so I'm a bit confused by this - if query needs to check X, Y and Z then it doesn't matter which of those are indexed, right?   Correct query should be constructed and then as we say below, it should be pointed out if there is no good index to support it.   (LMK if this is referencing something else that I missed)

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   - Check collection indexes before generating the query
+   - Avoid operators that prevent index usage: `$where`
+   - Do not use `$text` without a text index
+   - `$expr` should only be used when necessary (use sparingly)
+   - For compound indexes, use leftmost prefix when possible
```
asya999: same as above comment.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   - If no relevant index exists, mention this in your response (user may want to create one)
+2. **Avoid redundant operators** - Never add operators that are already implied by other conditions:
+   - Don't add `$exists: true` when you already have an equality check (e.g., `status: "active"` already implies the field exists)
```
asya999: not just equality check.  inequality check like $gt/$lt/etc also fit under this.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   - Don't add overlapping range conditions (e.g., don't use both `$gte: 0` and `$gt: -1`)
+   - Don't combine `$eq` with `$in` for the same field
```
asya999: not sure what this means...

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   - Each condition should add meaningful filtering that isn't already covered
+3. **Project only needed fields** - Reduce data transfer with projections
```
asya999, line 118: should mention to explicitly add `_id:0` when _id is not needed

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+4. **Validate field names** against the schema before using them
+5. **Handle edge cases** - Consider null values, missing fields, type mismatches
+5. **Use appropriate operators** - Choose the right MongoDB operator for the task:
+   - `$eq`, `$ne`, `$gt`, `$gte`, `$lt`, `$lte` for comparisons
+   - `$in`, `$nin` for matching values within arrays or lists
```
asya999: no, that's not what $in and $nin do - they are $eq or $ne that are $or'ed (in other words, `$in:[1,5]` is equivalent to `{$or:[{a:1},{a:5}]}` not "a is an array".

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   - `$and`, `$or`, `$not`, `$nor` for logical operations
+   - `$regex` for text pattern matching (prefer left-anchored patterns like `/^prefix/` when possible, as they can use indexes efficiently)
```
asya999: add "case sensitive" (or avoid case insensitive comparisons)

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   - `$exists` for field existence checks
```
asya999: Prefer `a:{$ne:null}` to `a:{$exists:false}`

skills/mongodb-natural-language-querying/SKILL.md
asya999: and it's not for type validation - it's for type *matching*

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Aggregation Pipeline Quality
+1. **Filter early** - Use `$match` as early as possible to reduce documents
+2. **Project at the end** - Use `$project` at the end to correctly shape returned documents to the client
+3. **Limit when possible** - Add `$limit` after `$sort` when appropriate
+4. **Use indexes** - Ensure `$match` and `$sort` stages can use indexes:
+   - Place `$match` stages at the beginning of the pipeline
+   - Initial `$match` and `$sort` stages can use indexes if they precede any stage that modifies documents
+   - Structure `$match` filters to align with available indexes
+   - Avoid `$unwind` or other transformations before `$match` when possible
```
asya999: this falls within putting $match early.  I would remove this.  Or rephrase as "minimize stages that transform documents before first $match"

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+5. **Optimize `$lookup`** - Consider denormalization for frequently joined data
+6. **Group efficiently** - Use accumulators appropriately: `$sum`, `$avg`, `$min`, `$max`, `$push`, `$addToSet`
```
asya999: grouping efficiently includes using $sort+$first instead of $min or $max.  Maybe this should be in the perf query and indexing skill?

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Error Prevention
+1. **Validate all field references** against the schema
+2. **Quote field names correctly** - Use dot notation for nested fields
+3. **Handle array fields properly** - Use `$elemMatch`, `$size`, `$all` as needed
```
asya999: is "as needed" too general?   Aren't we trying to encode into the skill when exactly it's needed?  One thing I see lacking all over is people checking for non-empty array as "does it exist/is type array" and that it's not equal to `[]` instead of more readable and efficient checking if first element exists...

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+4. **Escape special characters** in regex patterns
+5. **Check data types** - Ensure operations match field types from schema
```
asya999: operations or values?   `{a:5}` and `{a:"5"}` can both be either correct or not because of value not because of operator...

testing/mongodb-natural-language-querying/evals/evals.json
```diff
@@ -0,0 +1,250 @@
...
+      "name": "find-with-and-operator",
+      "prompt": "Return only the Plate IDs of Acura vehicles registered in New York",
+      "expected_output": "Find query with $and or implicit AND for vehicle make and state, project plate_id only",
+      "files": []
+    },
+    {
+      "id": 13,
+      "name": "find-with-non-english",
+      "prompt": "¿Qué alojamiento tiene el precio más bajo? devolver el número en un campo llamado \"precio\" en español",
+      "expected_output": "Find query with sort by price ascending, limit 1, project price field renamed",
```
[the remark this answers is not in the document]
asya999: Copilot is wrong.  As of 4.4 find projection and agg projection have the same capabilities.

testing/mongodb-natural-language-querying/evals/evals.json
```diff
@@ -0,0 +1,250 @@
...
+      "files": []
+    },
+    {
+      "id": 14,
+      "name": "find-with-regex-string-ops",
+      "prompt": "Write a query that does the following: find all of the parking incidents that occurred on any ave. Return all of the plate ids involved with their summons number and vehicle make and body type. Put the vehicle make and body type into lower case. No _id, sorted by the summons number lowest first.",
+      "expected_output": "Find query with regex for 'ave', projection with lowercase operations, sort by summons_number",
```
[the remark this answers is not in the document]
asya999: Again that’s wrong.

testing/mongodb-natural-language-querying/evals/evals.json
```diff
@@ -0,0 +1,250 @@
...
+      "prompt": "find all the movies released in 1983",
+      "expected_output": "Aggregation pipeline with $match on year 1983 (or preferably suggests find query)",
+      "files": []
+    },
+    {
+      "id": 17,
+      "name": "agg-filter-projection-sort-limit",
+      "prompt": "find three movies with alien in the title, show earliest movies first, only the _id, title and year",
+      "expected_output": "Aggregation with $match on title regex, $project, $sort, $limit",
+      "files": []
```
asya999: Not $search?

testing/mongodb-natural-language-querying/evals/evals.json
```diff
@@ -0,0 +1,250 @@
...
+      "prompt": "Return all the properties of type \"Hotel\" and with ratings lte 70",
+      "expected_output": "Aggregation with $match on nested property_type and ratings fields",
+      "files": []
+    },
+    {
+      "id": 21,
+      "name": "agg-group-sort-limit-project",
+      "prompt": "what is the bed count that occurs the most? return it in a field called bedCount (only return the bedCount field)",
+      "expected_output": "Aggregation with $group by beds, $count, $sort descending, $limit 1, $project to rename",
+      "files": []
```
asya999, line 135: Shouldn’t it just use $sortByCount?  (With limit and project)

testing/mongodb-natural-language-querying/evals/evals.json
```diff
@@ -0,0 +1,250 @@
...
+    },
+    {
+      "id": 22,
+      "name": "agg-group-sort-limit-project-2",
+      "prompt": "which host id has the most reviews across all listings? return it in only a field called hostId",
+      "expected_output": "Aggregation grouping by host_id, $sum reviews, $sort, $limit, $project",
+      "files": []
```
asya999, line 142: Is this right?  reviews field has number of reviews?

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   - If no relevant index exists, mention this in your response (user may want to create one)
+2. **Avoid redundant operators** - Never add operators that are already implied by other conditions:
+   - Don't add `$exists: true` when you already have an equality check (e.g., `status: "active"` already implies the field exists)
+   - Don't add overlapping range conditions (e.g., don't use both `$gte: 0` and `$gt: -1`)
+   - Don't combine `$eq` with `$in` for the same field
+   - Each condition should add meaningful filtering that isn't already covered
+3. **Project only needed fields** - Reduce data transfer with projections
+4. **Validate field names** against the schema before using them
+5. **Handle edge cases** - Consider null values, missing fields, type mismatches
+5. **Use appropriate operators** - Choose the right MongoDB operator for the task:
```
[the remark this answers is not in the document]
dacharyc: Tagging on to Copilot here, I was also about to flag this 😅

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   - For compound indexes, use leftmost prefix when possible
+   - If no relevant index exists, mention this in your response (user may want to create one)
+2. **Avoid redundant operators** - Never add operators that are already implied by other conditions:
+   - Don't add `$exists: true` when you already have an equality check (e.g., `status: "active"` already implies the field exists)
+   - Don't add overlapping range conditions (e.g., don't use both `$gte: 0` and `$gt: -1`)
+   - Don't combine `$eq` with `$in` for the same field
+   - Each condition should add meaningful filtering that isn't already covered
+3. **Project only needed fields** - Reduce data transfer with projections
+4. **Validate field names** against the schema before using them
+5. **Handle edge cases** - Consider null values, missing fields, type mismatches
```
dacharyc: We're instructing agents to handle edge cases here, but not advising on what actions to take. Do we think agents have enough info in their training data to handle them appropriately, or are there specific edge case handling patterns we want to call out?

testing/mongodb-natural-language-querying/evals/evals.json
```diff
@@ -0,0 +1,250 @@
...
+      "prompt": "group all the listings based on the amenities tags and return only count and tag name",
+      "expected_output": "Aggregation with $unwind on amenities, $group by amenity, $count",
+      "files": []
+    },
+    {
+      "id": 30,
+      "name": "agg-size-operator",
+      "prompt": "which listing has the most amenities? the resulting documents should only have the _id",
+      "expected_output": "Aggregation with $project using $size on amenities array, $sort, $limit, project _id only",
+      "files": []
```
asya999, line 198: Should/could be $addFields probably rather than $project first.

testing/mongodb-natural-language-querying/evals/evals.json
```diff
@@ -0,0 +1,250 @@
...
+    },
+    {
+      "id": 31,
+      "name": "agg-complex-regex-split",
+      "prompt": "What are the 5 most frequent words (case sensitive) used in movie titles in the 1980s and 1990s combined? Sorted first by frequency count then alphabetically. output fields count and word",
+      "expected_output": "Complex aggregation with $match on year range, $split on title, $unwind, $group with $sum, $sort",
```
asya999: $limit and second $sort missing here. 

Why is this called regex?

skills/mongodb-natural-language-querying/SKILL.md
dacharyc, line 118: We're saying to "project only needed fields" but haven't defined any criteria for an agent to determine what's needed. Is there anything we can/should say here about how the agent should decide what's "needed?"

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+
+**User Input:** "Find all active users over 25 years old, sorted by registration date"
+
+**Your Process:**
+1. Check schema for fields: `status`, `age`, `registrationDate` or similar
+2. Verify field types match the query requirements
+3. Check available indexes to optimize the query
+4. Generate query
+5. Suggest creating an index if no appropriate index exists for the query filters
+6. Suggest adding a limit if the collection is large or size is unknown
```
dacharyc: Is there some criteria we should give an agent for determining what "large" is?

testing/mongodb-natural-language-querying/mongodb-query-workspace/iteration-1/test-results.md
```diff
@@ -0,0 +1,311 @@
...
+<limit>1</limit>
+```
+
+**Grade:** ✅ **PASS**
+- ✅ Correctly interprets Spanish prompt
+- ✅ Correct sort ascending (lowest price first)
+- ✅ Correct limit 1
+- ✅ Correct field renaming to "precio"
+- ✅ Correctly excludes _id
+- Note: Expected format shows find query, but skill correctly chose aggregation for field renaming (which find cannot do)
```
asya999: Incorrect. This could be done in find.

testing/mongodb-natural-language-querying/mongodb-query-workspace/iteration-1/test-results.md
```diff
@@ -0,0 +1,311 @@
...
+---
+
+## Test 7: Complex Aggregation (Word Frequency)
+**Prompt:** What are the 5 most frequent words (case sensitive) used in movie titles in the 1980s and 1990s combined? Sorted first by frequency count then alphabetically. output fields count and word
+
+**Skill Output:**
+```json
+{
+  "aggregation": {
+    "pipeline": "[{ $match: { year: { $gte: 1980, $lte: 1999 } } }, { $project: { words: { $split: ['$title', ' '] } } }, { $unwind: '$words' }, { $group: { _id: '$words', count: { $sum: 1 } } }, { $sort: { count: -1, _id: 1 } }, { $limit: 5 }, { $project: { _id: 0, word: '$_id', count: 1 } }]"
```
asya999: It should be either $addFields instead of first $project or should exclude _ id.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+## Sample Document Usage
+
+Use sample documents to:
+- Understand actual data values and ranges
+- Identify field naming conventions (camelCase, snake_case, etc.)
+- Detect common patterns (e.g., status enums, category values)
+- Estimate cardinality for grouping operations
+- Validate that your query will work with real data
+
+## Common Pitfalls to Avoid
```
dacharyc: This section seems to largely duplicate guidance already stated in Error Prevention, Query Quality, and the context-gathering sections - is it needed here, or can we omit this section?

testing/mongodb-natural-language-querying/mongodb-query-workspace/iteration-1/test-results.md
```diff
@@ -0,0 +1,311 @@
...
+**Grade:** ✅ **PASS**
+- ✅ Correct $match for year range 1980-1999
+- ✅ Correct text splitting (uses $project instead of $addFields, both valid)
+- ✅ Correct $unwind to flatten words
+- ✅ Correct $group by word with count
+- ✅ Correct dual-level sort (count desc, word asc)
+- ✅ Correct limit 5
+- ✅ Correct output fields (word and count)
+- ✅ Case sensitive (no toLower)
+- Minor: Uses $project instead of $addFields for splitting, but functionally equivalent
```
asya999: Not exactly.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+## Size Limits
+
+Keep requests under 5MB:
+- If sample documents are too large, use fewer samples (minimum 1)
+- Limit to 4 sample documents by default
+- For very large documents, project only essential fields when sampling
+
+## Response Validation
```
dacharyc: This checklist seems to be restating advice from multiple earlier sections. Do we think there is a clear benefit to this format? If so, I'm wondering if the prior sections should be checklists themselves and maybe we can omit this since it seems duplicative?

asya999 (commented):
Find and agg can do same transformations. Other comments.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
+---
+name: mongodb-natural-language-querying
+description: Generate MongoDB queries (find) or aggregation pipelines using natural language, with collection schema context and sample documents. Use this skill whenever the user asks to write, create, or generate MongoDB queries, wants to search/filter/aggregate data in MongoDB, asks "how do I query...", needs help with query syntax, or discusses finding/filtering/grouping MongoDB documents. Generates queries that leverage available indexes. Also use for translating SQL-like requests to MongoDB syntax. Does NOT analyze or optimize existing queries - use mongodb-query-optimizer for that. Requires MongoDB MCP server.
+allowed-tools: mcp__mongodb__*, Read, Bash
```
dacharyc: Nit: `Read` and `Bash` are Claude-specific tool invocations. These won't be useful and could potentially prevent the skill from being used, or the relevant tools from being used, in other agent platforms where the tool names don't match, such as Cursor.

We can probably just omit these from the allowed tools list.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+---
+
+# MongoDB Natural Language Querying
+
+You are an expert MongoDB query generator. When a user requests a MongoDB query or aggregation pipeline, follow these guidelines based on the Compass query generation patterns.
```
dacharyc: We should probably scope-gate what types of `aggregation pipeline` activities we expect an agent to do with this skill, so an agent who needs to do some other type of aggregation operation we're not covering here, like outputting to another collection, can opt out of this skill early.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+
+**Use Find Query when:**
+- Simple filtering on one or more fields
+- Basic sorting and limiting
+- Field projection only
+- No data transformation needed
+
+**For Aggregation Pipelines**, generate an array of stage objects.
+
+**Use Aggregation Pipeline when the request requires:**
```
dacharyc, line 65: We might want to scope gate this, too. This skill covers read-oriented aggregation pipelines. If the request involves writing results to another collection (`$out`, `$merge`) or other write operations, this skill does not apply, and an agent applying our recommendations and best practices from here to one of those operation types could potentially drop data or documents that should be written/merged.

skills/mongodb-natural-language-querying/SKILL.md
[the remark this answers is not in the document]
dacharyc: Also flagging we might want to clarify the scope here, related to this - i.e.: we're saying here "Text search with scoring" as a reason to use an aggregation pipeline, which puts it explicitly in scope for this skill. But the only text search guidance in the skill is line 114: "Do not use `$text` without a text index." There's no mention of `$search` or `$vectorSearch` at all. Presumably that's intentional because we don't want to step on the search skill, but not being clear about the scope gate here is a problem because:
  - `$search` has completely different syntax, index requirements (Atlas Search indexes, not regular text indexes), and scoring mechanics than `$text`
  - `$search` must be the first stage in a pipeline, which conflicts with the "place `$match` at the beginning" advice (line 139)
  - An agent triggered by "search my documents for..." could produce a `$text`-based query when the user's deployment uses Atlas Search, or try to apply the skill's aggregation ordering rules to a `$search` pipeline where they don't hold
  - `$vectorSearch` is even further from anything the skill covers

The fix would be a scope-gate noting that Atlas Search (`$search`, `$vectorSearch`) queries are not covered by this skill.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+- [ ] Operators are used correctly for field types
+- [ ] Query syntax is valid MongoDB JSON
+- [ ] Query addresses the user's request
+- [ ] Query is optimized (filters early, projects when helpful)
+- [ ] Query can leverage available indexes (or note if no relevant index exists)
+- [ ] Response is properly formatted as JSON strings
+
+---
+
+## When invoked
```
dacharyc: Just wondering if we have found this section to be needed. It seems to repeat some of what we set up in the step-by-step instructions above. If we're trying to say here: "apply these steps in order" we might put that as guidance at the top of the skill instead of add it retroactively and repeat info here.

dacharyc (changes requested):
Hey @betsybutton - overall this Skill is in good shape; just a handful of comments related to some scope-gate concerns, cross-platform support, and a couple of places where I'm wondering if we can be more specific.

I think the scope gating and allowed tools suggestion is the only required change; the rest is on the subjective side.

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Got it! Thanks for clarifying

skills/mongodb-natural-language-querying/SKILL.md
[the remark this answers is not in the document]
betsybutton: Ah, good catch!

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Does "No complex document transformation needed" sound accurate to you?

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Okay yeah that's fair! When I was on Atlas, we would sometimes rethink what we wanted to query to avoid needing to create a new index. This is a bit nuanced and difficult to communicate through the skill, so it probably makes more sense to omit this part and focus on the other efficiency patterns.

skills/mongodb-natural-language-querying/SKILL.md
asya999: I guess I'm not sure why we are saying anything at all - why does it matter if agg is used everywhere?   If it can be reduced to find then it'll run internally as find.

skills/mongodb-natural-language-querying/SKILL.md
asya999: I'm guessing you would add fields that were already indexed to use an existing index but that's of very limited usefulness on most systems...

betsybutton: Yep, I'll just remove it.

skills/mongodb-natural-language-querying/SKILL.md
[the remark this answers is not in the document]
betsybutton: Opting to make the scopes cleaner cut; will remove this line & update other parts accordingly!

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Hmm do you think the skill should only generate aggregation pipelines then? I think find are slightly more readable if humans are reading the output v. agents.

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Will delete from here for now and discuss with David tomorrow

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Great point - proposing that we add the following to the `Query Quality` section and remove the bullet from the line you highlighted. 
```
7. **Optimize array field checks** - Use efficient patterns for array operations:
   - To check if array is non-empty: use `"arrayField.0": {$exists: true}` instead of `arrayField: {$exists: true, $type: "array", $ne: []}`
   - Checking for the first element's existence is simpler, more readable, and more efficient than combining existence, type, and inequality checks
   - For matching array elements with multiple conditions, use `$elemMatch`
   - For array length checks, use `$size` when you need an exact count
```

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: This should say value

testing/mongodb-natural-language-querying/evals/evals.json
betsybutton: Deleting this case given stricter boundaries - this should be search and it will be covered in Ayaan's tests

testing/mongodb-natural-language-querying/evals/evals.json
betsybutton: Thank you for noticing this mistake!

betsybutton: `number_of_reviews` is the name of the field

testing/mongodb-natural-language-querying/mongodb-query-workspace/iteration-1/test-results.md
betsybutton: I'm going to delete this file since it was just the results from the test run - but I applied your feedback to the test itself from your other comment for future runs

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Just going to delete this! It's handled elsewhere in the skill with more actionable info

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: If I were to expand here,  I would add something like "Decide appropriate fields to project based on given prompt context" which feels a bit tautological - the important guidance here is to remember to use projections in the first place

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Good callout; deleting!

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: I'll just delete this - we don't have enough info about their cluster in this skill yet to know what "large" is. There's no absolute threshold we can define.

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Deleting didn't seem to impact quality per evals, so I think it's fine to delete

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: I added something to the description to scope-gate the skill; do you think it needs to be mentioned here too?

dacharyc: I think your clarifications are probably good here so resolving!

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,208 @@
...
+- `project` - Field projection (optional)
+- `sort` - Sort specification (optional)
+- `skip` - Number of documents to skip (optional)
+- `limit` - Number of documents to return (optional)
+- `collation` - Collation specification (optional)
+
+**Use Find Query when:**
+- Simple filtering on one or more fields
+- Basic sorting and limiting
+- Field projection only
```
dacharyc: Asya pointed out that find can do the same field transformations as aggregation (projection language is identical since MongoDB 4.4), and the old bullet "No data transformation needed" was removed. But "Field projection only" on line 62 is still there, and it's now ambiguous. Does it mean "only simple include/exclude projections" or does it include computed expressions like `{ fullName: { $concat: ["$first", " ", "$last"] } }` which find also supports?

Meanwhile, the aggregation criteria still list "Computed fields or data reshaping" (line 69) as a reason to use aggregation. An agent could read these two sections and conclude it should use aggregation for a computed field projection that find handles just fine. This is a contradiction left over from the partial fix.

Worth investigating: Should "Field projection only" be clarified or removed? Or should the aggregation criteria for "Computed fields" be scoped to multi-stage reshaping that genuinely requires a pipeline?

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,208 @@
...
+- ❌ `{ age: { $gte: 25 } }` (object)
+
+For aggregation pipelines:
+- ✅ `"[{ $match: { status: 'active' } }]"` (string)
+- ❌ `[{ $match: { status: 'active' } }]` (array)
+
+## Best Practices
+
+### Query Quality
+1. **Generate correct queries** - Build queries that match user requirements, then check index coverage:
```
dacharyc, line 107: The rewrite from "structure queries for indexes" to "generate correct query first, then check index coverage" was a good change. But lines 111–112 now say:

 > - After generating the query, check if existing indexes can support it
 > - If no appropriate index exists, mention this in your response (user may want to create one)

And the example workflow (lines 188–189) reinforces this:

> 4. Check if available indexes can support the query
> 5. Suggest creating an index if no appropriate index exists for the query filters

The skill's updated description scope-gates with "Does NOT analyze or optimize existing queries — use mongodb-query-optimizer for that." But suggesting index creation is arguably optimization advice, not query generation. An agent could reasonably interpret this as: "generate the query AND do lightweight index analysis" - which overlaps with what the optimizer skill presumably does.

Worth investigating: Is "suggest creating an index" intentionally in-scope here, or should this skill stop at noting whether an existing index covers the query and defer index creation recommendations to the optimizer? The boundary test file only tests prompt routing, not whether the skill's content stays in its lane once triggered.

skills/mongodb-natural-language-querying/SKILL.md
```diff
@@ -0,0 +1,208 @@
...
+   - Each condition should add meaningful filtering that isn't already covered
+3. **Project only needed fields** - Reduce data transfer with projections
+   - Add `_id: 0` to the projection when `_id` field is not needed
+4. **Validate field names** against the schema before using them
+5. **Use appropriate operators** - Choose the right MongoDB operator for the task:
+   - `$eq`, `$ne`, `$gt`, `$gte`, `$lt`, `$lte` for comparisons
+   - `$in`, `$nin` for matching against a list of possible values (equivalent to multiple $eq/$ne conditions OR'ed together)
+   - `$and`, `$or`, `$not`, `$nor` for logical operations
+   - `$regex` for case sensitive text pattern matching (prefer left-anchored patterns like `/^prefix/` when possible, as they can use indexes efficiently)
+   - `$exists` for field existence checks (prefer `a: {$ne: null}` to `a: {$exists: false}` )
```
dacharyc: I'm wondering if we can clarify the guidance here?

These have different semantics:
- `{a: {$exists: false}}` — matches documents where the field a does not exist at all
- `{a: {$ne: null}}` — matches documents where a exists and is not null

A blanket "prefer X over Y" without noting the semantic difference could cause an agent to generate an incorrect query when the user actually wants to find documents where a field is missing (including cases where it's explicitly set to null). This isn't a scope-gating issue per se, but it's an accuracy concern introduced by applying Asya's feedback that could lead to wrong query results.

Worth investigating: Can you help us clarify intent here, @asya999 ? There may be a good reason (e.g., index behavior, common usage patterns) for the preference, but as written, an agent has no way to know when the preference doesn't apply.

dacharyc (approved):
Thanks for addressing my feedback, @betsybutton ! Overall I think we're in good shape here. Just a few small comments that we may want to investigate based on changes since my initial review, but I don't consider any of this blocking.

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: Thank you for noticing this inconsistency! I think that we can safely remove the "Field projection only" _and_ "Computed fields or data reshaping" bullets and rely on the "Multiple transformation stages" bullet for aggregations to cover the multi-stage/complex use case.

skills/mongodb-natural-language-querying/SKILL.md
betsybutton: This is intentionally in-scope. This is akin to Compass/DE's proactive performance advice, whereas the query optimizer skill is reactive and goes much further to justify/defend performance advice. I think we should leave this here for now, and consider merging the 2 skills in the future into the same skill. I'd be concerned about generating a query _without_ any very lightweight performance guidance.

dacharyc (approved):
Looks good to me! Clear to merge from my side whenever you're ready. ✅

PR #3, opened 2026-03-05
Add MCP setup skill MCP-426

nirinchev:

## Skill Information

**Skill Name:** `mongodb-mcp-setup`
**Skill Directory:** `skills/mongodb-mcp-setup`

## Use Case

Guides users through configuring the MCP server.

## Value Proposition

Simplifies the credentials config for the MCP server.

## Validation Prompts

TODO:

1. Can you help me configure the MongoDB MCP server?
2. [User] List my mongodb databases
    [Agent] You need to connect first ... do you want me to help you setup the MCP server with credentials to connect.
    [User] Yes
3. Can you help me ensure the mongodb mcp server only has read access to my cluster?
4. I'm trying to use the MongoDB MCP tools to manage my Atlas clusters but I keep getting authentication errors - help me resolve that.
5. Looks like the mongodb mcp server is configured to use my local testing environment, but I want to now connect to a staging cluster - how do I do that?

## Author Self-Validation

- [x] Verified structure and format requirements
- [x] Checked content quality and novelty
- [x] Validated all links and file references
- [x] (Optional but recommended) Ran [`skill-validator`](https://github.com/dacharyc/skill-validator) locally

## SME Review

**SME:** @blva


skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+
+The MongoDB MCP server requires authentication credentials to work. Users have two options:
+
+1. **Connection String** (Option A): Direct connection to a specific MongoDB cluster
+   - Simpler setup
+   - Single cluster access
+   - Requires `MDB_MCP_CONNECTION_STRING` environment variable
+
+2. **Service Account Credentials** (Option B): MongoDB Atlas Admin API access
+   - More complex setup
```
blva: This might discourage the setup steps leading to Atlas users using connection strings, we should say it's a setup for MongoDB Atlas Users

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+description: Guide users through configuring the MongoDB MCP server. Use this skill when a user has the MongoDB MCP server installed but hasn't configured the required environment variables, or when they ask about connecting to MongoDB/Atlas and don't have the credentials set up. Also trigger immediately after the user installs the MongoDB plugin. This skill helps users choose between direct database connection (connection string) or Atlas Admin API access (service account credentials), then walks them through the setup process.
+---
+
+# MongoDB MCP Server Setup
+
+This skill guides users through configuring the MongoDB MCP server for use with an agentic client.
+
+## Overview
+
+The MongoDB MCP server requires authentication credentials to work. Users have two options:
```
blva: there's a third option, no configuration and using atlas local to create a local instance

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+
+1. **Connection String** (Option A): Direct connection to a specific MongoDB cluster
+   - Simpler setup
+   - Single cluster access
+   - Requires `MDB_MCP_CONNECTION_STRING` environment variable
+
+2. **Service Account Credentials** (Option B): MongoDB Atlas Admin API access
```
blva, line 18: are links OK for skills, if so we should probably link:
1) service accounts - https://www.mongodb.com/docs/atlas/api/api-authentication/#service-accounts-overview 
2) user roles https://www.mongodb.com/docs/atlas/reference/user-roles/

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -9,19 +9,27 @@ This skill guides users through configuring the MongoDB MCP server for use with
...
-   - Single cluster access
+   - Quick setup for single cluster access
+   - Requires manual connection string with database credentials
    - Requires `MDB_MCP_CONNECTION_STRING` environment variable
 
 2. **Service Account Credentials** (Option B): MongoDB Atlas Admin API access
-   - More complex setup
-   - Access to Atlas Admin API
+   - **Recommended for Atlas users** - manages authentication internally and simplifies data access
+   - Full access to Atlas Admin API
```
blva: not full access, it depends on their service account roles so I would say scoped access or just access

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -9,19 +9,27 @@ This skill guides users through configuring the MongoDB MCP server for use with
...
    - Ability to connect to any cluster dynamically using the `atlas-connect-cluster` tool
+   - No need to create DB Users or manage their credentials
```
blva: it creates a db user under the hood in some cases, it might even create via tool call right?

nirinchev: Right - we do create db users under the hood, this was intended more to contrast the active steps a human must take - e.g. with a connection string, you'd need to manually create the db user and plug the credentials into the connection string, whereas with service accounts, this is more automated, so humans wouldn't need to fiddle with that. But I can tweak the wording to be more precise, you're right.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+
+#### 5a.3: Add Environment Variables
+
+Add the environment variables to the appropriate profile file. Add a comment to make it clear what these are for:
+
+For Connection String (Option A):
+```bash
+# MongoDB MCP Server Configuration
+export MDB_MCP_CONNECTION_STRING="<value>"
+```
```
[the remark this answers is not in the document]
nirinchev: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/mongodb/agent-skills/pull/3#discussion_r2918363828)

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+### For Option C (Atlas Local):
+
+1. **Ready to use**: No restart or configuration needed - the MCP server works with Atlas Local out of the box!
+
+2. **Next steps**: They can now:
+   - Create a local deployment with `atlas-local-create-deployment`
+   - List existing deployments with `atlas-local-list-deployments`
+   - Once connected to a deployment, use all standard database operations (find, insert, update, delete, aggregate, etc.)
+
+## Important Notes
```
dacharyc, line 361: The `review-skill` skill flagged this as a potential source of token inefficiency, suggesting this is repeating troubleshooting guidance already covered in Step 6.

Regarding the security line, it seems to already be covered in two other places:

- Permissions (600): Step 5a.4 (line 284-285) already sets chmod 600 on the profile file during Interactive Mode.
- Never commit credentials: The SETUP.md template in Step 5b already includes a "Security reminder" section (lines 359-360) with "Never commit credentials to git" and "Keep secrets in your shell profile only."

So we may be able to safely omit this section entirely.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+- **Troubleshooting**: If the MCP server still doesn't work after restart:
+  - Verify the environment variables are set in a fresh terminal: `env | grep MDB_MCP`
+  - Check that the client was fully restarted (not just reloaded)
+  - Verify the credentials are valid by testing them directly (connection string by connecting with `mongosh`, service account credentials by making an Atlas API call)
+  - If using read-only mode, verify `MDB_MCP_READ_ONLY` is set to `true`
+  - Check the client's MCP server logs for error messages
+
+## Error Handling
+
+Be prepared for common issues:
```
dacharyc: This also seems like a candidate for consolidation or removal. Seems like a lot of this guidance is already covered elsewhere, with one exception (according to Claude):

  Looking at the Error Handling section (lines 416-422):

  1. "Bash permission denied" — Already handled inline at Step 1 (line 52: "If Bash access is denied: Skip to Documentation Mode"), Step 5 (line 219: "If Bash is denied at any point, switch to Documentation Mode"), and Step 5a.1 (line 232: "If Bash is denied: Switch immediately to Documentation Mode").
  2. "Invalid connection string format" — Already handled inline at Step 3a.2 (lines 102-105), which covers
  validation and what to do if it doesn't look valid.
  3. "Profile file doesn't exist" — Already handled inline at Step 5a.3 (line 278: "If the file doesn't exist,
   create it") and Step 5b's template includes instructions for finding/creating the profile.
  4. "Permission denied on profile file" — This is the one item that's not covered inline. Step 5a.4 sets
  permissions proactively but doesn't handle the case where the agent can't write to an existing profile file due to permissions.
  5. "Variables not loading" — Already covered in the Troubleshooting sections of both Step 6 (lines 407-412) and the SETUP.md template (lines 352-355).
  6. "Service account credentials invalid" — Already covered in Step 6 (line 411: "Verify the credentials are
  valid...").

  So 5 of 6 items are duplicates of inline handling. The only net-new guidance is "Permission denied on
  profile file." That one could be inlined into Step 5a.3 (where the agent writes to the profile) and Step
  5b's documentation. After that, the entire Error Handling section could be removed, which would improve
  token efficiency.

(Assuming any of this guidance does not generally apply to more than one step, in which case calling it out separately here and referring to it from inline may be the most token-efficient approach.)

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+Offer to open this URL in their browser to make it easier.
+
+### 3b.3: Guide Through Key Steps
+
+While they work through the documentation, remind them of the key steps:
+
+1. **Navigate to MongoDB Atlas** - Go to cloud.mongodb.com and sign in
+2. **Access Organization Settings** - Find the organization where they want to create the service account
+3. **Create Service Account** - Go to Access Manager → Service Accounts → Create Service Account
+4. **Set Permissions** - Grant appropriate permissions (they'll need at least Organization Member or Project Owner for most operations)
```
dacharyc: The "appropriate permissions" phrase here is doing a lot of heavy lifting, and agents may benefit from more specific guidance. The skill says "they'll need at least Organization Member or Project Owner for most operations" but doesn't specify which permissions map to which use cases. An agent can't advise a user on what to select without knowing: if they only need data access, what's the minimum? If they need cluster management, what's required? This is open to interpretation and could lead to over- or under-provisioning.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+
+Without proper API Access List configuration, they'll encounter errors when trying to use Atlas Admin API tools.
+
+### 3b.4: Collect Credentials
+
+Once they've completed the Atlas setup, ask them to provide:
+- Client ID
+- Client Secret
+
+Use the agent's user-input capability with appropriate fields for these credentials.
```
dacharyc: Step 3a has an explicit validation step (3a.2) for connection strings. Step 3b has no equivalent. The skill doesn't say whether the agent should validate the format of the Client ID/Secret (e.g., expected length, character set) or just accept whatever the user provides. An agent might or might not attempt validation. If we can say more here about whether validation is needed, and if so, what type - or explicitly state that validation is not needed - it could shortcut any interpretation logic that may result in different outcomes.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+**Try to detect** the user's current shell:
+```bash
+echo $SHELL
+```
+
+**If Bash is denied:** Switch immediately to Documentation Mode (Step 5b).
+
+Based on the shell, determine the appropriate profile file to update:
+
+| Shell | Profile File |
```
dacharyc: The table covers zsh, bash, and fish, with a ~/.profile fallback. But the skill doesn't define what to do if the agent detects a shell not in the table (e.g., nushell, tcsh). Should it use the fallback silently, or warn the user that their shell isn't fully supported?

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+export MDB_MCP_API_CLIENT_ID="<value>"
+export MDB_MCP_API_CLIENT_SECRET="<value>"
+```
+
+**If the user chose read-only access** (Step 4), add this additional line:
+```bash
+export MDB_MCP_READ_ONLY="true"
+```
+
+Use available file-editing capabilities to append these lines to the profile file. If the file doesn't exist, create it with file-creation capabilities.
```
dacharyc: This is implicit about how to write to the file. Should the agent append to the end? Insert after existing exports? Insert before or after a specific marker? If the profile file has conditional blocks or sourcing logic, appending to the end might not work as expected. The skill could specify "append to the end of the file" explicitly.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+**Partial Configuration Handling:**
+- If user wants to add read-only mode to existing setup (has auth but no `MDB_MCP_READ_ONLY`), skip to Step 4
+- If user wants to switch authentication methods, explain they should remove the old variables from their shell profile first, then proceed with Steps 2-5
+- If user wants to update existing credentials, offer to update the configuration in their shell profile
+
+**Important**: If the user is asking to perform an Atlas Admin API action (like managing clusters, creating database users, getting performance advisor recommendations) but only has `MDB_MCP_CONNECTION_STRING` configured, explain that they need service account credentials for Atlas Admin API access and offer to help them set it up.
+
+## Step 2: Present Configuration Options
+
+If no valid configuration exists, present the options to the user and help them understand which one suits their needs:
```
dacharyc: We're missing guidance on what to do if the user can't decide what suits their needs. The skill says to present options and "use the agent's interactive question/choice capability" but doesn't specify a default recommendation or what to do if the user is unsure. Step 3b's overview says Service Account is "Recommended for Atlas users" but Step 2 doesn't carry that forward as a default suggestion when the user is uncertain.

Of these, step 1 (permissions) and step 4 (file editing behavior) are the most likely to cause an agent to do the wrong thing. The others are edge cases but still worth tightening.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+```bash
+env | grep -E "MDB_MCP_(CONNECTION_STRING|API_CLIENT_ID|API_CLIENT_SECRET|READ_ONLY)"
+```
+
+**If Bash access is denied:** Skip to Documentation Mode (see Step 5b). You can't check existing configuration interactively, so proceed with creating documentation files.
+
+**Interpretation (if Bash succeeded):**
+- If `MDB_MCP_CONNECTION_STRING` is set, the user has connection string auth configured
+- If both `MDB_MCP_API_CLIENT_ID` and `MDB_MCP_API_CLIENT_SECRET` are set, the user has service account auth configured
+- If the user has either valid configuration, exit gracefully unless they explicitly need Atlas Admin API access but only have the connection string configured
```
dacharyc: When existing valid configuration is found, the skill says to "exit gracefully" but doesn't specify what that looks like. Should the agent confirm what it found and say nothing more? Should it list the configured variables and their values? Should it ask the user if they want to change anything? Different agents will interpret "exit gracefully" differently.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+
+Before starting the setup, check if the user already has the required environment variables configured.
+
+**Try to run** this command to check for existing configuration:
+
+```bash
+env | grep -E "MDB_MCP_(CONNECTION_STRING|API_CLIENT_ID|API_CLIENT_SECRET|READ_ONLY)"
+```
+
+**If Bash access is denied:** Skip to Documentation Mode (see Step 5b). You can't check existing configuration interactively, so proceed with creating documentation files.
```
dacharyc: If Bash is denied at Step 1, the skill says to skip to Step 5b (Documentation Mode). But this skips Steps 2-4 entirely — the user never gets asked which auth option they want, never provides credentials, and never chooses read-only vs read-write. The SETUP.md template in 5b covers all three options generically, but the agent hasn't gathered any user-specific context. Should the agent still walk through Steps 2-4 conversationally (without Bash) before creating the documentation?

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+
+**Interpretation (if Bash succeeded):**
+- If `MDB_MCP_CONNECTION_STRING` is set, the user has connection string auth configured
+- If both `MDB_MCP_API_CLIENT_ID` and `MDB_MCP_API_CLIENT_SECRET` are set, the user has service account auth configured
```
dacharyc: The skill defines what to do when both API credentials are set, or when the connection string is set, but doesn't explicitly handle the case where only one of the two service account variables is present. Is this partial config? Should the agent treat it as unconfigured and start from Step 2, or ask the user to do something about the missing secret?

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+Ensure the profile file has appropriate permissions (especially important for files containing secrets):
+```bash
+chmod 600 ~/.zshrc  # or whatever profile file was used
+```
+
+#### 5a.5: Verify Configuration
+
+Test that the variables are set correctly by sourcing the profile:
+```bash
+source ~/.zshrc  # or appropriate profile file
```
dacharyc: The verification step shows source ~/.zshrc but source isn't valid in fish shell (which uses source with different semantics) or other non-POSIX shells. Given that Step 5a.1 explicitly supports fish, the verification step should account for it — e.g., . ~/.config/fish/config.fish or a fish-specific equivalent.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+## Step 3c: Atlas Local Setup
+
+If the user chooses Option C:
+
+### 3c.1: Check Docker Installation
+
+Verify that Docker is installed and running:
+
+```bash
+docker --version
```
dacharyc: The skill checks docker --version which verifies installation but not that the Docker daemon is actually running. An agent could pass this check and then have atlas-local-create-deployment fail later. A check like `docker info` or `docker ps` would confirm the daemon is active.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -0,0 +1,423 @@
...
+- On macOS, prefer `~/.bash_profile` unless you can confirm `~/.bashrc` is sourced.
+- Avoid writing only to `~/.bashrc` when the environment uses login-shell startup files.
+
+**For Windows users** (if you detect Windows OS):
+- PowerShell: Use the PowerShell profile (check `$PROFILE` variable)
+- Inform them they may need to create the profile if it doesn't exist
+
+#### 5a.2: Check for Existing Configuration
+
+Before adding new environment variables, check if these variables are already defined in the profile file. If they are, offer to update them rather than adding duplicates.
```
dacharyc: Step 5a.2 says to "check if these variables are already defined" and "offer to update them rather than adding duplicates," but doesn't specify how to update — should the agent replace the existing line in-place, comment out the old one and append a new one, or something else? Commenting out preserves history; replacing in-place is cleaner. The choice affects the user's profile file, so it should be explicit.

dacharyc (commented):
Overall this looks very helpful - the scope gating seems clear, the steps are self-contained, and the information is presented in a novel way. Suggestions here are mostly about tightening logic and token efficiency, and/or reducing the number of cases where an agent may need to interpret what to do, and therefore could behave in unexpected or undesired ways depending on the interpretation of a given agent in a given skill run.

dacharyc (approved):
Thanks for making these updates, @nirinchev  - everything looks reasonable to me, and I think this will be helpful for folks trying to use our tooling. ✅

PR #5, opened 2026-03-05
feat(skills): mongodb-connection MCP-419

RaschidJFR:

## Skill Information

**Skill Name:** mongodb-connection
**Skill Directory:** `skills/mongodb-connection`

## Use Case

MongoDB connection configuration is deceptively complex. Developers often copy connection code from tutorials or Stack Overflow without understanding how pool sizes, timeouts, and connection patterns affect their specific environment. This leads to common production issues like connection pool exhaustion in serverless functions, timeout errors under load, and performance degradation in high-traffic applications. This skill addresses these challenges by providing context-aware guidance for MongoDB connection management across all officially supported driver languages.

## Value Proposition

This skill prevents costly misconfigurations by enforcing the principle of "context before configuration." Instead of providing arbitrary parameter values, it asks about your environment (serverless, long-running, traffic patterns) and provides tailored recommendations. This approach:

Saves debugging time - Catch connection issues before they reach production
Improves performance - Right-sized pools and timeouts for your use case
Multi-language support - Works with Node.js, Python, Java, Go, C#, Ruby, PHP, and more
Production-ready patterns - Optimized for Lambda, Express APIs, and high-traffic applications
Educational - Learn MongoDB connection best practices while building

## Special Considerations

None.

## Validation Prompts

1. "Create an AWS Lambda function that connects to MongoDB"
2. "My MongoDB connections are timing out with ECONNREFUSED errors, help me troubleshoot"
3. "Optimize my Express API's MongoDB connection pool for high traffic"
4. "Review this MongoDB client configuration for a serverless function"
5. "Add a new API endpoint to receive IoT sensor data and store it in MongoDB"

## Author Self-Validation

- [x] Verified structure and format requirements
- [x] Checked content quality and novelty
- [x] Validated all links and file references
- [x] (Optional but recommended) Ran [`skill-validator`](https://github.com/dacharyc/skill-validator) locally

## SME Review

**SME:** @rozza 

## Additional Context

~~Initial results look promising. More comprehensive testing is required. **Do not merge yet!**~~
Ready to merge.


skills/mongodb-connection/references/language-patterns.md
```diff
@@ -0,0 +1,417 @@
+# Language-Specific Connection Patterns
+
+This reference provides language-specific patterns and considerations for MongoDB connection management. Consult this when working with a specific driver or when users ask about best practices for their language.
```
Meir017: Would it make sense to have a skill per language? This way information about other languages won't clutter the llms context

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,440 @@
...
+    socketTimeoutMS: 30000,      // 30s - prevent hanging queries
+    maxIdleTimeMS: 600000,       // 10 min - stable server lifecycle
+    waitQueueTimeoutMS: 3000,    // 3s - quick feedback on saturation
+});
+```
+
+## Troubleshooting Connection Issues
+
+### Pool Exhaustion
+**Symptoms**: `MongoWaitQueueTimeoutException`, increased latency, operations waiting for connections
```
rozza: In Java this would be a `MongoTimeoutException`

rozza (commented):
Added a note about the Java driver.

skills/mongodb-connection/SKILL.md
RaschidJFR: Actually, I didn't find `MongoWaitQueueTimeoutException` in [any of the languages](https://github.com/search?q=org%3Amongodb+MongoWaitQueueTimeoutException&type=code).

Node uses [MongoWaitQueueTimeoutError](https://github.com/search?q=org%3Amongodb+MongoWaitQueueTimeout&type=code) and Java, [MongoTimeoutException](https://github.com/search?q=org%3Amongodb+MongoTimeoutException&type=code). Do we need to look for the specific error/exception per language?

skills/mongodb-connection/references/language-patterns.md
RaschidJFR: Having a skill per language, though ideal, would be hard to maintain initially. But maybe removing the code examples is not a terrible idea.

skills/mongodb-connection/SKILL.md
rozza: Lots use [`WaitQueueTimeoutError`](
https://github.com/search?q=org%3Amongodb+WaitQueueTimeoutError+-language%3AText+-language%3Ayaml+-language%3Ajson&type=code)

With the message: "Timed out while checking out a connection from connection pool".

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,295 @@
...
+
+- **`connectTimeoutMS`** (5-10s): Fail fast on unreachable servers
+- **`socketTimeoutMS`** (30s OLTP, 60-300s OLAP): Prevent hanging queries. Always non-zero.
+- **`maxIdleTimeMS`** (10-30s serverless, 5-10min long-running): Balance reuse vs cleanup
+- **`waitQueueTimeoutMS`** (2-5s): Fail fast when exhausted
+
+## Troubleshooting Connection Issues
+
+### Pool Exhaustion
+**Symptoms**: `MongoWaitQueueTimeoutError` or `MongoTimeoutException`, increased latency, operations waiting
```
rozza:
```suggestion
**Symptoms**: `MongoWaitQueueTimeoutError`, `WaitQueueTimeoutError` or `MongoTimeoutException`, increased latency, operations waiting
```

skills/mongodb-connection/SKILL.md
rozza: Added a suggestion to improve.

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,493 @@
...
+- [Connection Churn Diagnosis](#connection-churn-diagnosis)
+- [Setting Up Monitoring](#setting-up-monitoring)
+
+---
+
+## Driver-Level Metrics (Client-Side)
+
+Modern MongoDB drivers expose connection pool telemetry, providing a client-side view of connection health. Access methods vary by driver:
+- **Node.js**: Event listeners (`client.on('connectionPoolCreated', ...)`)
+- **Python**: `client.get_server_pool_stats()`
```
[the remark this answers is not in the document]
rozza: This is a good catch of some AI hallucination (perhaps it existed in a legacy pymongo?). This should read:

```
- **Python**: Event listeners (`pymongo.monitoring.register(listener)`)
```

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,493 @@
...
+- **Java**: `ConnectionPoolListener` interfaces
+- **Go**: Monitoring through driver configuration
```
rozza: Recommend we remove the langauge examples and just state:

```suggestion
Modern MongoDB drivers expose connection pool telemetry, providing a client-side view of connection health. Access methods by configuring a connection pool listener in the client settings / options.
```

rozza (changes requested):

skills/mongodb-connection/references/monitoring-guide.md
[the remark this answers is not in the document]
rozza: I added a recommendation below to remove the language examples.

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: Language examples removed.

skills/mongodb-connection/SKILL.md
RaschidJFR: added

skills/mongodb-connection/references/language-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+**When to increase**: You can increase `maxPoolSize` beyond 5 for Node.js, but:
+- Monitor for connection timeout issues under load
+- If you need more concurrency, consider multiple `MongoClient` instances (different Node processes/instances)
+- Alternative: Use a more scalable driver (Python, Java) if you hit Node's scaling limits
+
+### Best Practices
+- **Singleton pattern**: Create one `MongoClient` instance and export it
+- **Efficient with smaller pools**: Event loop multiplexes many operations over fewer connections (5-20 typically sufficient)
+- **Module-level initialization**: For serverless, initialize outside the handler
+- **Don't blindly increase pool size**: Node.js efficiency comes from its async model, not connection count
```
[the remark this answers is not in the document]
RaschidJFR: section updated

skills/mongodb-connection/references/monitoring-guide.md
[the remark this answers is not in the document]
RaschidJFR: Language examples removed.

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,261 @@
...
+
+**Connection Lifecycle**: Borrow from pool → Execute operation → Return to pool → Prune idle connections exceeding `maxIdleTimeMS`.
+
+**The wait queue is your canary.** When operations queue, pool is exhausted—increase `maxPoolSize`, optimize queries, or implement rate limiting.
+
+**Synchronous vs. Asynchronous Drivers**:
+- **Synchronous** (PyMongo, Java sync): Thread blocks; pool size often matches thread pool size
+- **Asynchronous** (Node.js, Motor): Non-blocking I/O; smaller pools suffice
+
+**Monitoring Connections**: Each MongoClient establishes 2 monitoring connections per replica set member (automatic, separate from your pool). Formula: `Total = (minPoolSize + 2) × replica members × app instances`. Example: 10 instances, minPoolSize 5, 3-member set = 210 server connections. Always account for this when planning capacity.
```
RaschidJFR: @rozza

skills/mongodb-connection/references/language-patterns.md
```diff
@@ -131,19 +131,17 @@ The PHP extension manages connection pooling at the process level. In traditiona
...
-- ✅ Other drivers' default of 100 handles moderate traffic for sync drivers
+- For most applications, the default of 100 is a reasonable starting point
+- Async drivers (Node.js, Motor) can handle high concurrency with fewer connections due to non-blocking I/O
+- Sync drivers' default of 100 handles moderate traffic well
 
 **When to adjust**:
 - Increase if you observe sustained connection pool exhaustion (wait queue growth, >80% utilization)
 - Decrease for low-traffic applications to reduce resource footprint
+- Decrease for serverless environments (3-10 per function instance)
 - **Don't change without reason**—defaults are based on extensive real-world usage patterns
```
RaschidJFR: @rozza

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -160,9 +160,9 @@ Server-side metrics provide the MongoDB server's perspective on connection usage
 
 **What it is**: MongoDB's WiredTiger storage engine uses a ticket-based concurrency control system. Tickets represent slots for concurrent read and write operations. When all tickets are in use, additional operations must wait.
 
-**Default ticket counts** (MongoDB 3.6+):
-- Read tickets: 128
-- Write tickets: 128
+**Default ticket counts**:
+  - **MongoDB 3.6 - 6.x**: Fixed at 128 read tickets and 128 write tickets
+  - **MongoDB 7.0+**: Starts at 128 but uses dynamic adjustment algorithm (may vary based on workload)
```
RaschidJFR: @rozza

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: @rozza

RaschidJFR (commented):
@rozza Can you help me verify some sections I recently updated? I tagged you in the comments.

skills/mongodb-connection/references/language-patterns.md
```diff
@@ -0,0 +1,166 @@
...
+- Horizontal scaling (more application instances with smaller pools each)
+- If you hit Node's scaling limits, consider multi-threaded drivers (Python, Java, Go)
+
+**Note for legacy applications**: If using MongoDB Node.js Driver 3.x or Mongoose 5.x (pre-2021), the default was pool size 5. Consider upgrading to modern versions for better defaults and performance
+
+---
+
+## Python
+
+### PyMongo (Synchronous)
```
rozza: Pymongo supports sync and async now.
https://pymongo.readthedocs.io/en/stable/api/pymongo/asynchronous/index.html

skills/mongodb-connection/references/language-patterns.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Java
+
+### Key Characteristics
+- **Both sync and async APIs**: Choose based on application architecture
+- **Thread-per-request common**: Traditional servlet containers
+- **Reactive Streams support**: For reactive frameworks
+
+### Best Practices
+- **Spring Boot**: Configure via `MongoClientSettings` bean
```
rozza: No need to mention Spring Boot here.

skills/mongodb-connection/references/language-patterns.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Key Characteristics
+- **Context-based timeouts**: Idiomatic Go pattern for cancellation and timeouts
+- **Goroutine-safe**: Client can be shared across goroutines
+- **Default pool size**: `maxPoolSize` defaults to 100 connections
+
+### Best Practices
+- **Prefer context timeouts over driver timeouts**: Use `context.WithTimeout` for operation-level control
+- **Default pool usually sufficient**: 100 connections handles most workloads
+- **Package-level initialization**: Share client across application
```
rozza: Alot of these best practises are the same across drivers.
I think the language used should be normalized.

eg these are the same suggestions with different verbage across languages:

*  **One client per application**: Singleton pattern via dependency injection
*  **Package-level initialization**: Share client across application
* **One client instance per application lifecycle**: MongoClient is expensive to create
* **Global client**: Initialize once, use throughout application

rozza (commented):
Think its look really good.

Great catch on the node default pool size - seems the models have out of date data.

Just a couple of comments of feedback.

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,396 @@
...
+
+---
+
+### WiredTiger Tickets Available
+
+**What it is**: MongoDB's WiredTiger storage engine uses a ticket-based concurrency control system. Tickets represent slots for concurrent read and write operations. When all tickets are in use, additional operations must wait.
+
+**Default ticket counts**:
+  - **MongoDB 3.6 - 6.x**: Fixed at 128 read tickets and 128 write tickets
+  - **MongoDB 7.0+**: Starts at 128 but uses dynamic adjustment algorithm (may vary based on workload)
```
rozza: Not sure this is correct. 128 appears to be the max. 
https://www.mongodb.com/docs/manual/core/wiredtiger/#transaction--read-and-write--concurrency

However, it may be lower.

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,396 @@
...
+
+**Important diagnostic pattern**:
+- **Client shows wait queue + Server has tickets available**: Increase client `maxPoolSize`
+- **Client pool healthy + Server tickets exhausted**: Server-side bottleneck; optimize queries or scale server
+- **Both exhausted**: Need both client pool increase AND server capacity/optimization
+
+**How to check**: Query `db.serverStatus().wiredTiger.concurrentTransactions` to see:
+- `read.available` / `write.available` - Tickets currently available
+- `read.out` / `write.out` - Tickets currently in use
+- `read.totalTickets` / `write.totalTickets` - Total ticket count (typically 128 each)
```
rozza:
```suggestion
- `read.totalTickets` / `write.totalTickets` - Total ticket count
```

rozza (approved):
LGTM!

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### 3.1 Key Principle: Every Parameter Must Be Justified
+
+When you suggest configuration, explain WHY each parameter has its specific value based on the context you gathered. Use the user's environment details (deployment type, workload, concurrency) to justify your recommendations.
+
+#### 3.2 Configuration Examples by Scenario
+
+**These are reference templates—adapt them to the user's specific context from Phase 1.** Each scenario below applies when the user described that environment during context gathering.
+
+**Language-specific implementations**: For Python, Java, Go, C#, Ruby, or PHP, see `references/language-patterns.md` for complete code examples and driver-specific patterns.
```
dacharyc: Looks like there aren't any code examples in the referenced file, so we may want to remove this reference to them.
```suggestion
**Language-specific implementations**: For Python, Java, Go, C#, Ruby, or PHP, see `references/language-patterns.md` for driver-specific patterns.
```

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+- Ensure `minPoolSize` > 0 for traffic spikes
+- Network compression for high-latency (>50ms): `compressors: ['snappy', 'zlib']`
+- Nearest read preference for geo-distributed setups
+
+### Server-Side Connection Limits
+Total connections = instances × maxPoolSize × replica members. Monitor `connections.current` to avoid hitting limits.
+
+## Language-Specific Considerations
+
+Configuration examples above are Node.js-based. For Python, Java, Go, C#, Ruby, or PHP: consult `references/language-patterns.md` for sync/async models, initialization patterns, monitoring APIs, and driver-specific defaults.
```
dacharyc: Since we're not actually showing any examples, we may want to avoid using that term here.

Also, can we be specific about which things above are Node.js based? Is it the parameter names we're providing, or availability/implementation details in each Driver, or something else? I'm seeing Driver-specific patterns in the referenced file, but nothing like what we're showing above, so I'm having trouble finding the connection between the things above that might be Node.js-based and their analogs in the language-patterns.md file for the other Drivers.

Also, we mention here that users will find driver-specific defaults, but the only default supplied in the referenced file is the 100-connection `maxPoolSize`. Since that's the same across all drivers, it seems misleading to characterize info in the related file as providing "driver-specific defaults."

We also say "monitoring APIs" here, but this is the only monitoring-related content in the referenced file:

```
### Monitoring Access
Most drivers provide:
- **Event listeners**: Subscribe to connection pool events
- **Statistics APIs**: Query current pool state
- **Logging**: Enable debug logging for troubleshooting
```

I wouldn't characterize that as "monitoring APIs", nor does it seem particularly helpful or to cover anything beyond what's probably already in the LLM's base training data.

```suggestion
Configuration scenarios above are Node.js-based. For Python, Java, Go, C#, Ruby, or PHP: consult `references/language-patterns.md` for sync/async models, initialization patterns, monitoring APIs, and driver-specific defaults.
```

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+- **Current metrics** (if available): Ops/sec, average latency
+
+#### For Troubleshooting (When Errors Reported)
+- **Error message**: Complete error (ECONNREFUSED, SocketTimeout, MongoWaitQueueTimeoutException, etc.)
+- **When**: Cold starts? Under load? Intermittent? Consistent?
+- **Current config**: Existing pool settings?
+- **Pool metrics**: Connections in use? Wait queue?
+- **Connectivity test**: Connects via mongo shell from same environment?
+
+Ask follow-up questions if responses are vague.
```
dacharyc: Can we provide more clarity here for what the agent should consider "vague" and what types of follow up questions to ask? For example, something like this: "If the user does not specify deployment type, concurrency level, or workload pattern, ask for those details before proceeding."

In other words, what is the minimum information an agent needs to proceed past this step? We need to make it clear what's required and how to elicit relevant details.

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Phase 2: Analysis and Diagnosis
+
+Analyze whether this is a client config issue or infrastructure problem.
+
+**Infrastructure Issues (Out of Scope)** - redirect appropriately:
```
dacharyc: Maybe I am missing it somewhere in this PR, but we're instructing agents to analyze whether this is a client config issue or infrastructure problem, and not giving agents any details about how to identify infrastructure issues. Can we provide a concrete decision tree or diagnostic sequence to help agents make this determination?

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+- DNS/SRV resolution failures, network/VPC blocking, IP not whitelisted, TLS cert issues, auth mechanism mismatches
+
+**Client Configuration Issues (Your Territory)**:
+- Pool exhaustion, inappropriate timeouts, poor reuse patterns, suboptimal sizing, missing serverless caching, connection churn
+
+When identifying infrastructure issues, explain: "This appears to be a [DNS/VPC/IP] issue rather than client config. It's outside the scope of the client configuration skill, but here's how to resolve: [guidance/docs link]."
```
dacharyc: If we intend an agent to provide docs links or guidance, we need to give the agent that info to pass along. Can we add relevant guidance or links to docs where agents can find the info to pass to the user?

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Pool Exhaustion
+**Symptoms**: `MongoWaitQueueTimeoutError`, `WaitQueueTimeoutError` or `MongoTimeoutException`, increased latency, operations waiting
+
+**Diagnosis**: Current `maxPoolSize`? Concurrent operations? Long-running queries or unclosed cursors?
+
+**Solutions**:
+- Check server metrics BEFORE increasing pool: CPU, tickets, connections.current
+- **Increase `maxPoolSize`** when: Wait queue + server has capacity (available tickets, <70% CPU)
+- **Don't increase** when: Server at capacity (tickets exhausted, high CPU)—optimize queries instead
```
dacharyc: Here, we're saying "optimize queries instead" - but that's outside the scope of this skill. We might want to add an instruction here to make it clear the agent should not attempt to be helpful and optimize the query as part of this skill workflow, something like: "Advise the user that query optimization is needed and is outside the scope of connection configuration."

skills/mongodb-connection/references/language-patterns.md
```diff
@@ -0,0 +1,171 @@
+# Language-Specific Connection Patterns
```
dacharyc: The information in this file seems to be very general and to largely restate information that is available in our docs/LLM training data. 

We could probably remove the non-novel information and consolidate this into something much smaller. It would also be good to test coding agents on connection patterns in these Drivers with and without the skill loaded. Have you found that agents are unable to adhere to these patterns without this guidance, and that adding this info improves the code gen/agent patterns?

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
+# MongoDB Connection Monitoring Guide
+
+This reference provides detailed guidance on monitoring connection pool health, interpreting metrics, and taking action based on what you observe. Consult this when users need to verify their configuration is working or troubleshoot connection-related issues.
+
+---
+
+## Driver-Level Metrics (Client-Side)
+
+Modern MongoDB drivers expose connection pool telemetry, providing a client-side view of connection health. Access methods vary by driver:
```
dacharyc: Have you tested monitoring-related code gen tasks with and without the skill loaded? I'm wondering if agents _need_ this info or if it's readily available to them in their base training data. I mainly ask because when I was testing Agent Skills, I observed a degradation in agent output when presented with minimal or partial API reference details like this. 

If we _need_ to show these, we might need to show more detail here - and if we _don't_ need to show them, we should omit them to avoid different types of failure modes.

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
...
+- **Node.js**: Event listeners (`client.on('connectionPoolCreated', ...)`)
+- **Python**: `client.get_server_pool_stats()`
+- **Java**: `ConnectionPoolListener` interfaces
+- **Go**: Monitoring through driver configuration
+
+### Connections Created (`totalCreated`)
```
dacharyc: For this and the following sections, do we need to provide guidance to agents about _how_ to watch for these things? Is the answer for all of these something like "use the client-side connection pool telemetry described above for your Driver"?

It's unclear to me from reading this file if we expect the agent to already know/have established patterns for this, and how the information we're providing here fits into those patterns. I'm concerned that we may either be missing providing relevant "how" details, scope gates, decision trees, and other tools for agents to understand the data flow and take action - or else if those details aren't _needed_ - have we been able to validate that providing _these_ patterns improves agent outputs?

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
...
+
+**What it is**: The total number of connections the pool has established since initialization.
+
+**What to watch for**: Rapid increases indicate connection churn due to network issues or misconfiguration.
```
dacharyc: Instead of saying "Rapid increases" here without providing the agent the info needed to assess what is rapid, and then restating down below a variant - "Rapid growth" - can move that info up here? i.e.:


```suggestion
**What to watch for**: Rapid increases (+100 connections/hour in steady state) indicate connection churn due to network issues or misconfiguration.
```

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
...
+
+**Healthy pattern**: Gradual increase during application startup as the pool warms up, then relatively stable. You should see increases mainly when:
+- Application restarts
+- Pool size is increased
+- Network disruptions force reconnections
+
+**Troubleshooting**:
+- **Rapid growth** (+100 connections/hour in steady state): Indicates connection churn. Check:
```
dacharyc: We're giving the agent a list of things to check here, but no info about how to check them. What criteria should the agent be looking for? What do we mean when we say "too aggressive?" Is there a way for the agent to check that the application is not creating new clients repeatedly - certain types of patterns to watch out for, debugging details we would suggest?

I have no idea if an agent knows how to do these things without instruction, so this is probably an area where we need more specific "how" info or testing of how an agent handles these scenarios without it.

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
...
+### Connections In-Use
+
+**What it is**: The number of connections currently borrowed from the pool and serving application requests.
+
+**What to watch for**: Consistently high values approaching `maxPoolSize` signal potential pool exhaustion.
+
+**Healthy pattern**: Fluctuates with application traffic while maintaining headroom. Should correlate with request volume.
+
+**Action thresholds**:
+- **Sustained >80% of maxPoolSize**: Increase `maxPoolSize` by 20-30%
```
dacharyc, line 63: These threshholds are specific and that's good! But we could provide some more details, like:

- Sustained over what period of time?
- Consistently 100% over what period of time? At all? Over 5 minutes? Over an hour?
- When we say "high wait queue times" - what constitutes "high?" What should the agent expect wait queue times to be when the connection pool is healthy?

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
...
+- **Consistently 100%**: Pool is definitely exhausted; immediate action needed
+- **High percentage with high wait queue times**: Clear sign of undersized pool
+
+**Diagnosis questions**:
+- Does it correlate with traffic spikes?
```
dacharyc: Diagnostic questions are good! But can we tie these back to specific mitigations?

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
...
+- Are operations taking longer than expected to complete?
+- Are there any long-running queries holding connections?
+
+---
+
+### Connections Available (Idle)
```
dacharyc: Maybe this is a silly question, but I'm asking since I'm not familiar with how we normally communicate about these things. Is "available connections" really different than "connections in use?" We're talking about them like they're different things here, but it seems to me like these are both facets of a "connection usage" bucket, and you can derive this as a corollary of the section above.

I'm asking because LLMs have a tendency to attribute to or intuit meaning from distinctions like this, and if there really isn't a distinction, we may accidentally convince one that there is - i.e. that it should take different actions for the same underlying problem, not enough connections in the pool.

In other words, if these are really just different sides of the same thing, we can create cognitive load on the part of the agent by _presenting_ them as though they're different, and that can have downstream consequences on agent performance and whether agents take the correct actions.

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
...
+
+**Action thresholds**:
+- **Always zero during traffic**: Pool is too small; connections are never released
+- **Very low during normal load**: Consider increasing `maxPoolSize` or `minPoolSize`
+
+---
+
+### Wait Queue Size
+
+**What it is**: The number of operations currently waiting for an available connection because the pool is at capacity.
```
dacharyc, line 89: Apologies again if this is a silly q, but I had a quick look around the docs, and it wasn't obvious to me if there is a way to know the size of the wait queue. We're speaking here like there is, but I didn't spot anything obvious. Is this something we need to provide instructions for?

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,397 @@
...
+**Action thresholds**:
+- **>50ms consistently**: Pool is under pressure; investigate sizing
+- **>100ms**: Immediate action required; users experiencing degraded performance
+- **Spikes to >waitQueueTimeoutMS**: Users seeing timeout errors
+
+---
+
+## Server-Level Metrics (MongoDB-Side)
+
+Server-side metrics provide the MongoDB server's perspective on connection usage. Access via:
```
dacharyc: These seem like three different abstractions, and it's unclear to me what we expect the agent to do with this info.

I don't think there's a way for agents to access MongoDB Atlas monitoring dashboards (if there is, we should provide details about it), and it's unclear to me whether agents have a way to understand whether an integration with a monitoring platform is available or how to interface with it. That leaves `db.adminCommand()` in the agent's purview, but as far as I know, from the Driver side, executing admin commands is a one-time operation - not a listener that would facilitate monitoring. So if there is a monitoring pattern we expect agents to use, we probably need to provide details about what that should be.

dacharyc (commented):
Hey @RaschidJFR - back to you with what I've gotten through so far. I've spent a couple of hours with this, and my feedback is mainly about a lack of specificity throughout. I've done a quick skim through the rest of the `monitoring-guide.md` and if I continued to review, it would be a lot more similar comments.

What we have here seems to look useful on the surface, but it's unclear to me if it provides enough information to effectively improve agent outputs. I'd love to learn more about how you identified the areas of focus for this skill, and what types of performance differences you observed when testing without the skill and with the skill.

skills/mongodb-connection/references/language-patterns.md
RaschidJFR: Updated.
- General patterns have moved to the bottom in a section for cross-language best practices
- Now keeping only language specific information in each language section

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: There are no hard guidelines or absolute values on this in the literature, but this has been enough to produce consistent recommendations when troubleshooting.

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: There are no hard guidelines or absolute values on this in the literature, but this has been enough to produce consistent recommendations when troubleshooting.

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: Removed diagnostic questions.

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: reworked section for clarity and included the relevant events to subscribe to for each metric.

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: Added instructions on how to calculate it.

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: Removed reference to Atlas dashboard

skills/mongodb-connection/SKILL.md
RaschidJFR: Added examples to distinguish between each type of issue.

skills/mongodb-connection/SKILL.md
RaschidJFR: Good catch. This was vague. Updated.

skills/mongodb-connection/SKILL.md
RaschidJFR: Not sure if advising the user that something is outside the scope of a skill and stopping is useful. That would limit the proactivity of the agent to resolve the problem (even if outside the scope of the skill) as it would normally do, constraining it to act only within the limits of a skill. Probably the user wouldn't care either if it's in or out of the scope as long as the agent acts.

skills/mongodb-connection/SKILL.md
RaschidJFR: Good catch. That line was outdated. Fixed, thx.

dacharyc (approved):
Hey @RaschidJFR  - after examining this skill in more detail, I don't spot any big quality or inconsistency issues here. It would be nice to have some language-specific prompts in the evals, just to validate that the language patterns are actually helpful and provide good guidance, but I'm considering that a nice-to-have at this point. I don't want to block merging on it given that you have found the skill improves at least some cases in evals.

Thank you for iterating on this!

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,224 @@
...
+### WiredTiger Tickets Available
+
+**What it is**: MongoDB's WiredTiger storage engine uses a ticket-based concurrency control system. Tickets represent slots for concurrent read and write operations. When all tickets are in use, additional operations must wait.
+
+**How to check**: Query `db.serverStatus().wiredTiger.concurrentTransactions` to see:
+- `read.available` / `write.available` - Tickets currently available
+- `read.out` / `write.out` - Tickets currently in use
+- `read.totalTickets` / `write.totalTickets` - Total ticket count
+
+**What to watch for**: Low available tickets indicate the server is at maximum concurrency capacity, regardless of connection availability.
```
asya999: Our docs explicitly say that this is not the case as of 7.0 and new(er) dynamic algorithm (from our docs: 
Note
A low value of [available](https://www.mongodb.com/docs/manual/reference/command/serverStatus/#mongodb-serverstatus-serverstatus.available) in [queues.execution](https://www.mongodb.com/docs/manual/reference/command/serverStatus/#mongodb-serverstatus-serverstatus.queues.execution) does not indicate a cluster overload. Use the number of queued read and write tickets as an indication of cluster overload.

Shouldn't we be pointing to queued (high) rather than available (low) as indicator?

asya999 (commented):
comment about available tickets

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,224 @@
...
+- **Tickets available = 0 frequently**: Server is at capacity; operations will queue even if connections are available
+- **Persistent ticket exhaustion**: Consider query optimization, indexing improvements, or scaling server tier
+- **Tickets exhausted + low CPU**: May indicate lock contention or slow I/O
+
+**Important diagnostic pattern**:
+- **Client shows wait queue + Server has tickets available**: Increase client `maxPoolSize`
+- **Client pool healthy + Server tickets exhausted**: Server-side bottleneck; optimize queries or scale server
+- **Both exhausted**: Need both client pool increase AND server capacity/optimization
+
+**See manual**: [WiredTiger concurrentTransactions](https://www.mongodb.com/docs/manual/reference/command/serverStatus/#mongodb-serverstatus-serverstatus.wiredTiger.concurrentTransactions)
```
asya999: looks like a bad hash - just goes to the top level of that page

doug-rodgers-mongodb (commented):
This looks great from a security perspective. The only recommendation is to explicitly call out something like the following:

“When working with connection strings, never ask the user to paste actual passwords or secrets. Prefer environment variables or redacted URIs (e.g., mongodb+srv://<user>:<password>@cluster/...). Do not echo full URIs or credentials into responses.”

skills/mongodb-connection/references/monitoring-guide.md
RaschidJFR: Thanks for this. Removing the WiredTiger section since available tickets are no longer an indicator after 7.0+ dynamic value and WiredTiger's metrics do not guarantee server health.

testing/mongodb-connection/evals/evals.json
```diff
@@ -0,0 +1,96 @@
...
+      "id": 2,
+      "prompt": "Using MongoDB, give me a minimal snippet for new endpoint for a high-frequency trading system with 100k MAU. This endpoint is to place ask/bid orders. Provide a suitable client configuration.",
+      "expected_output": "Agent should ask about specific requirements for the app. Should then provide a code snippet that demonstrates the connection setup with suitable parameters.",
+      "expectations": [
+        "Client initialization should follow singleton, module-level, or cached, client pattern",
+        "Verifies or assumes OLTP workload",
+        "maxPoolSize must be more than 50",
+        "minPoolSize should be between 10 and 50",
+        "socketTimeout and connectTimeout should be greater than zero and less than a minute",
+        "Must justify the connection parameters based on the context of the task"
```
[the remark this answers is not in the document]
RaschidJFR: Leaving vague on purpose to not restrict to a particular language (prop names vary across driver languages)

skills/mongodb-connection/references/monitoring-guide.md
```diff
@@ -0,0 +1,196 @@
...
+**Why this matters**: If `waitQueueTimeoutMS` is reached, users see errors.
+
+---
+
+#### Wait Queue Time
+
+**What it is**: The duration operations spend waiting for connections to become available.
+
+**Events** – Calculate duration: `(checked out time) - (checkout started time)`
+- `ConnectionCheckoutStartedEvent` - record timestamp when entering queue
```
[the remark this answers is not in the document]
RaschidJFR: ignore

skills/mongodb-connection/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Pool Exhaustion
+When operations queue, pool is exhausted.
+
+**Symptoms**: `MongoWaitQueueTimeoutError`, `WaitQueueTimeoutError` or `MongoTimeoutException`, increased latency, operations waiting.
+
+**Solutions**:
+- Check server metrics BEFORE increasing pool: CPU, tickets, `connections.current`
+- **Increase `maxPoolSize`** when: Wait queue has operations waiting (size > 0) + server shows low utilization (available tickets, low CPU)
+- **Don't increase** when: Server at capacity (tickets exhausted, high CPU). Suggest query optimization.
```
asya999: this still references tickets from before ... should be made consistent

asya999 (changes requested):
need to remove ticket reference from SKILL.md

skills/mongodb-connection/SKILL.md
RaschidJFR: removed

dacharyc (approved):
LGTM on my side - thank you for all your work on this, @RaschidJFR ! ✅

asya999 (approved):
LGTM

PR #7, opened 2026-03-06
Schema design skill MCP-420

addaleax:

## Skill Information

**Skill Name:** MongoDB Schema Design
**Skill Directory:** `skills/mongodb-schema-design`

## Use Case

We want to be able to give our users accurate and helpful schema modeling guidance. This can be in the context of general performance improvement investigations or when initially designing schemas for example.

## Value Proposition

We know existing LLMs do not perform as well as they could when presented with questions around MongoDB schema design, and we do want to remedy that gap.

## Special Considerations

None (yet). We will likely want to have more MCP server integration in the future, esp. around gathering workload information.

## Validation Prompts

  1. Design a MongoDB data model for a dog walking app.
  2. I'm migrating our e-commerce platform from PostgreSQL to MongoDB. We have tables for users, orders, order_items, products, and inventory. How should I structure these as MongoDB collections?                                                      
  3. Our social media app is experiencing slow queries when loading user profiles with their posts and comments. Each user can have thousands of posts. Should I embed posts in the user document or keep them in a separate collection?                                                    
  4. We're hitting the 16MB document limit in our analytics collection where we store user activity events as an array. What's the best way to fix this? 
  5. I need to model a product catalog where products can have different attributes depending on their category (clothing has sizes/colors, electronics have specs, books have ISBNs). What's the MongoDB pattern for this?

## Author Self-Validation

- [ ] Verified structure and format requirements
- [ ] Tested skill with LLMs using the five prompts above
- [ ] Checked content quality and novelty
- [ ] Validated all links and file references
- [ ] (Optional but recommended) Ran [`skill-validator`](https://github.com/dacharyc/skill-validator) locally

## SME Review

**SME:** @asya999, @johnlpage

## Additional Context

The main gaps currently are:

1. a relatively low novelty score of the main SKILL.md file – which feels tricky to address, since yes, it mainly re-states facts about what is already public available information about MongoDB data modeling, but at the same time it seems to genuinely help the LLM in a "hammer the point home" way where repeating the core principles is helpful.
2. The relatively large size of the references files – we'll want to cut down on the number and the content size without losing too much information


skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
```diff
@@ -0,0 +1,153 @@
...
+db.reviews.createIndex({ productId: 1 })  // Critical for $lookup
+
+// Use pipeline $lookup for filtered joins
+db.products.aggregate([
+  { $match: { _id: productId } },
+  { $lookup: {
+      from: "reviews",
+      let: { pid: "$_id" },
+      pipeline: [
+        { $match: { $expr: { $eq: ["$productId", "$$pid"] } } },
```
asya999: Should use local/foreign syntax on join equality rather than $expr

skills/mongodb-schema-design/references/antipattern-massive-arrays.md
```diff
@@ -0,0 +1,131 @@
...
+    },
+    $inc: { commentCount: 1 }
+  }
+)
+// Simultaneously insert into comments collection
+db.comments.insertOne({ postId: "post123", ...newComment })
+```
+
+**Alternative ($slice without separate collection):**
+
```
asya999: I would think the simpler case is all comments in separate collection.  I’m not sure we usually advise discarding older comments.

skills/mongodb-schema-design/references/antipattern-schema-drift.md
```diff
@@ -0,0 +1,196 @@
...
+title: Prevent Schema Drift
+impact: CRITICAL
+impactDescription: "Prevents application errors and query failures from inconsistent schemas"
+tags: schema, anti-pattern, validation, consistency, data-quality, atlas-suggestion
+---
+
+## Prevent Schema Drift
+
+**Schema drift—when documents in the same collection have inconsistent structures—causes application errors and query inconsistencies.** MongoDB's flexibility is a feature, but undisciplined field additions lead to code that must handle many shapes. Use schema validation to prevent drift before it happens.
+
```
asya999: Maybe add “unexpected” somewhere - as in different schema can be fine by design but mistake if it’s done unintentionally

skills/mongodb-schema-design/references/antipattern-schema-drift.md
```diff
@@ -0,0 +1,196 @@
...
+{ _id: 4, firstName: "Dave", lastName: "Jones", emails: ["dave@ex.com", "d@work.com"] }
+
+// Application code becomes defensive nightmare
+function getUserEmail(user) {
+  if (user.email) return user.email
+  if (user.emails) return user.emails[0]
+  throw new Error("No email found")  // Crashes on some documents
+}
+
+// Queries fail silently
```
asya999: Maybe not fail but fail to return all documents?

skills/mongodb-schema-design/references/antipattern-unbounded-arrays.md
```diff
@@ -0,0 +1,103 @@
+---
+title: Avoid Unbounded Arrays
```
asya999: Is this significantly enough different from massive arrays?

asya999 (commented):
A few comments on the high value ones

skills/mongodb-schema-design/references/antipattern-massive-arrays.md
addaleax: Yup, dropped this subsection

skills/mongodb-schema-design/references/antipattern-unbounded-arrays.md
addaleax: I don't know, but it's always seemed like an issue for me when we did the work around this anti-pattern in Compass that we were talking about "unbounded" arrays but really meant "large" arrays (since we weren't actually observing whether they were potentially growing or not)

asya999: I've been thinking about this.  Unbounded means conceptually it never stops growing (so a list of my orders is an example) but it's true if it's likely to stay small (a list of my shipping addresses) then it's not "as big" of a problem.

skills/mongodb-schema-design/references/antipattern-bloated-documents.md
```diff
@@ -0,0 +1,77 @@
...
+db.products.aggregate([
+  { $project: {
+    total: { $bsonSize: "$$ROOT" },
+    imagesSize: { $bsonSize: { $ifNull: ["$images", {}] } },
+    reviewsSize: { $bsonSize: { $ifNull: ["$reviews", {}] } }
+  }},
+  { $group: {
+    _id: null,
+    avgTotal: { $avg: "$total" },
+    avgImages: { $avg: "$imagesSize" },
```
asya999: if we are worried about this, I'd wonder if $max would also be valuable - either $max of $bsonSize or $max of $size of array.  But LLM cannot run this (can it?) - do we provide this so that it can suggest this as a verification method to the user?

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
```diff
@@ -0,0 +1,82 @@
...
+
+- **Data changes frequently and independently**: If brand logos change daily, denormalization creates update overhead.
+- **Rarely-accessed data**: Don't embed review details if only a small fraction of product views load reviews.
+- **Many-to-many with high cardinality**: Avoid embedding large or fast-growing relationship sets.
+- **Analytics queries**: Batch jobs can afford $lookup latency; real-time queries cannot.
+
+## Verify with
+
+```javascript
+// Find pipelines with multiple $lookup stages
```
asya999, line 56: wouldn't it be better to add an `$or` `filter` for specifically agg with `$lookup`?  I know it's a little ugly given our command format but it would literally find all `$lookup` pipelines, rather than just the ones that exceed an arbitrary threshold...

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
```diff
@@ -0,0 +1,82 @@
...
+    $elemMatch: { "$lookup": { $exists: true } }
+  }
+}).sort({ millis: -1 })
+
+// Check if $lookup foreign fields are indexed
+db.reviews.aggregate([
+  { $indexStats: {} }
+])
+// Look for "productId_1" with high ops - good
+// Missing index = every $lookup is a collection scan
```
asya999: couldn't it be a compound index with productId prefix? or in cases of multiple join conditions, productId wouldn't even need to be first..

skills/mongodb-schema-design/references/antipattern-schema-drift.md
```diff
@@ -0,0 +1,186 @@
...
+{ _id: 1, name: "Alice", email: "alice@ex.com" }
+
+// Version 2 (2021) - added phone
+{ _id: 2, name: "Bob", email: "bob@ex.com", phone: "555-1234" }
+
+// Version 3 (2022) - restructured name
+{ _id: 3, firstName: "Carol", lastName: "Smith", email: "carol@ex.com" }
+
+// Version 4 (2023) - email is now array
+{ _id: 4, firstName: "Dave", lastName: "Jones", emails: ["dave@ex.com", "d@work.com"] }
```
asya999: the issue isn't that it's an array (since equality matches scalar and array containing the same) it's that the name of the field now changed to `emails` from `email`

skills/mongodb-schema-design/references/antipattern-schema-drift.md
```diff
@@ -0,0 +1,186 @@
...
+
+// Application code becomes defensive nightmare
+function getUserEmail(user) {
+  if (user.email) return user.email
+  if (user.emails) return user.emails[0]
+  throw new Error("No email found")  // Crashes on some documents
+}
+
+// Queries fail silently or unintentionally return all documents
+db.users.find({ email: "test@ex.com" })  // Misses users with emails[] array
```
asya999: I'd say with name of field `email`

skills/mongodb-schema-design/references/antipattern-schema-drift.md
```diff
@@ -0,0 +1,186 @@
...
+
+```javascript
+// Define and enforce consistent schema
+db.createCollection("users", {
+  validator: {
+    $jsonSchema: {
+      bsonType: "object",
+      required: ["email", "profile"],
+      properties: {
+        email: {
```
asya999: why wouldn't we allow an array (with the correct name) since some people could have multiple email addresses?

skills/mongodb-schema-design/references/antipattern-schema-drift.md
```diff
@@ -0,0 +1,186 @@
...
+// Missing validator means schema drift risk is higher
+
+// Primary check: find documents that do NOT match current validator
+if (validator) {
+  db.users.find({ $nor: [validator] }).limit(20)
+  db.users.countDocuments({ $nor: [validator] })
+}
+
+// Optional heavy check for maintenance windows:
+// validate can be slow and can take an exclusive lock on the collection.
```
asya999: this looks for corruption, is this related to validator and schema?  I'm concerned this doesn't include explanation what it's useful for.

skills/mongodb-schema-design/references/fundamental-schema-validation.md
```diff
@@ -0,0 +1,134 @@
...
+  validationLevel: info?.options?.validationLevel,
+  validationAction: info?.options?.validationAction,
+  validator: info?.options?.validator
+})
+
+// Primary compliance check: find documents that do NOT match validator
+const validator = info?.options?.validator
+db.users.find({ $nor: [validator] })
+
+// Optional heavyweight check (slow, can block due to a collection lock):
```
asya999: again, not sure why this is here - also I'd prefer we don't add to confusion about blocking and locking.

skills/mongodb-schema-design/references/relationship-one-to-many.md
```diff
@@ -0,0 +1,108 @@
...
+
+**When to use One-to-Many references:**
+
+| Scenario | Example | Why Reference |
+|----------|---------|---------------|
+| Unbounded children | Publisher → Books | Could have 100,000+ books |
+| Large child documents | User → Orders | Orders have line items, addresses |
+| Independent queries | Department → Employees | Query employees directly |
+| Different lifecycles | Author → Articles | Archive articles separately |
+| Frequent child updates | Post → Comments | Adding comments shouldn't lock post |
```
asya999: we don't lock on updates.

skills/mongodb-schema-design/references/antipattern-schema-drift.md
```diff
@@ -0,0 +1,186 @@
...
+db.runCommand({
+  collMod: "users",
+  validator: {
+    $jsonSchema: {
+      properties: {
+        schemaVersion: { enum: [1, 2] }  // Accept both during migration
+      }
+    }
+  },
+  validationLevel: "moderate"  // Don't block existing invalid docs
```
asya999: the word "block" is ambiguous - can we say instead "error on" (everywhere)?

skills/mongodb-schema-design/references/relationship-one-to-many.md
addaleax: Dropped

skills/mongodb-schema-design/references/antipattern-schema-drift.md
addaleax: Yes, let's drop this

skills/mongodb-schema-design/references/fundamental-schema-validation.md
addaleax: Ditto

skills/mongodb-schema-design/references/antipattern-schema-drift.md
addaleax: Added a comment here, I feel like explicitly adding an $anyOf example with the two would just increase file size without really providing more information

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
addaleax: I've shortened this a bit to just say "Look for index supporting the query in result", I think an LLM should be able to fill the gaps here

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
addaleax: I haven't been able to get a `filter` for `db.setProfilingLevel()` to work that doesn't choke on looking for `$lookup` because of the $-prefixed key ... but yes, that feels somewhat preferable

skills/mongodb-schema-design/references/antipattern-bloated-documents.md
addaleax: Added $max 👍 

Agent should be able to run aggregation pipelines via the MCP server, so yeah, these should actually be usable as automated validation/verification as well

skills/mongodb-schema-design/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+- Seeing Atlas Schema Suggestions or Performance Advisor warnings
+- Hitting the 16MB document limit
+- Adding schema validation to existing collections
+
+## Rule Categories by Priority
+
+| Priority | Category | Impact | Prefix | Rules |
+|----------|----------|--------|--------|-------|
+| 1 | Schema Anti-Patterns | CRITICAL | `antipattern-` | 6 |
+| 2 | Schema Fundamentals | HIGH | `fundamental-` | 4 |
```
dacharyc: This row labels all "Schema Fundamentals" as HIGH impact, but individual reference files vary: fundamental-16mb-awareness.md is CRITICAL, fundamental-schema-validation.md is MEDIUM. The fundamentals mismatch could potentially confuse an agent about priority.

skills/mongodb-schema-design/references/fundamental-schema-validation.md
dacharyc, line 1: There is some content overlap between reference files:
(1) fundamental-schema-validation.md and validation-json-schema.md both demonstrate `$jsonSchema` validators with similar code;
(2) antipattern-bloated-documents.md and pattern-subset.md both cover hot/cold data separation;
(3) antipattern-unbounded-arrays.md and fundamental-16mb-awareness.md overlap on unbounded arrays and the 16MB limit.

Given the 32K+ token total, I'm wondering if there's a way to break any of these overlapping details out to their own reference files and consolidating to reduce reference token load. Maybe the answer is no, or the overlap is trivial enough that it wouldn't actually be that beneficial, but just wanted to call this out as a possible opportunity.

skills/mongodb-schema-design/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+
+A core way to implement this philosophy is the fact that MongoDB exposes **flexible schemas**. This means you can have different fields in different documents, and even different structures. This allows you to model data in the way that best fits your access patterns, without being constrained by a rigid schema. For example, if different documents have different sets of fields, that is perfectly fine as long as it serves your application's needs. You can also use schema validation to enforce certain rules while still allowing for flexibility.
+
+Another implication of the key principle is that information about the expected read and write workload becomes very relevant to schema design. If pieces of information from different entities are often queried or updated together, that means that prioritizing co-location of that data in the same document can lead to significant performance benefits. On the other hand, if certain pieces of information are rarely accessed together, it may make sense to store them separately to avoid loading more data than necessary.
+
+#### Schema Fundamentals Summary
```
dacharyc, line 64: The LLM-as-judge scoring identified this as A) partially duplicative of some of the reference files, and B) largely information that is already known in training data/public docs.

I'm wondering if we can drop this, or if you found in testing that this was needed?

skills/mongodb-schema-design/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+- [pattern-schema-versioning](references/pattern-schema-versioning.md) - Evolve schemas safely with version fields
+- [pattern-subset](references/pattern-subset.md) - Store hot data in main doc, cold data elsewhere
+- [pattern-time-series-collections](references/pattern-time-series-collections.md) - Use native time series collections when available
+
+### 5. Schema Validation (MEDIUM) - 2 rules
+
+- [validation-json-schema](references/validation-json-schema.md) - Validate data types and structure at database level
+- [validation-action-levels](references/validation-action-levels.md) - Choose warn vs error mode for validation
+
+## Key Principle
```
dacharyc, line 54: Is any of this novel information, and/or did you find in testing that the skill needed this? It seems to be publicly available information so I'm wondering if we can drop it.

dacharyc (commented):
Found a handful of small things in SKILL.md, but there's a bigger-picture thing I think we need to address with this skill before publication. We're missing any sort of routing information for an agent to decide _which_ of the reference files it should read. The list of links includes things like: `Decision framework for relationships` and `Embrace documents, avoid SQL patterns` - which describes what's _in_ each reference, but not how an agent should choose which reference files to load.

The "How To Use" section says "Read individual reference files for detailed explanations and code examples" - which implies the agent may need to read all of them to figure out which ones are relevant. At 30 reference files and over 30k tokens, just loading this skill and following an instruction to read the reference files will use a quarter of a frontier model's context window, and could take minutes. And given research like [Lost in the Middle](https://arxiv.org/abs/2307.03172) and similar, which explores that LLMs have trouble retrieving relevant details in the middle of long context, we can't assume that an LLM will actually retrieve relevant details for schema modeling decision making if it _does_ load all of the reference files.

I think we need to find ways to help agents self-select a subset of this reference content before we publish this. Even if the reference files themselves are great, we need to provide better guardrails to help agents benefit from the curation.

skills/mongodb-schema-design/SKILL.md
addaleax: @dacharyc Right, this was one thing I noticed during eval case testing – for the really essential core stuff, we got better results when there was an extra summary here in SKILL.md itself. Without it, the agent tended to drift towards SQL/relational-style approaches, even when loading the relevant reference files.

I did notice that this reduces the novelty score as reported by the skill checker (it's expected, because, yes, this is publicly available information), but since the ultimate goal here is to help LLMs provide better answers, I'd assume that eval case results are a stronger indicator than other scores.

skills/mongodb-schema-design/SKILL.md
addaleax: (Basically same answer as below)

skills/mongodb-schema-design/references/fundamental-schema-validation.md
addaleax: No, it's a good callout – we know we want to reduce token count anyway. I've pushed a few changes to consolidate these files. I assume the downside of this is that we're ending up with larger reference files, so we may end up including more information that isn't as relevant as we'd like it to be (if the LLM picks one of the original files only instead of both, that is).

skills/mongodb-schema-design/SKILL.md
addaleax: Yeah, I feel like the impact column hasn't exactly been super helpful anyway, and you're right that it should be more important to describe to the agent when to consult which reference file rather than marking some files as "CRITICAL" (unless they really are "always-load" – but then they should probably be part of the SKILL.md file anyway, right?)

skills/mongodb-schema-design/SKILL.md
dacharyc: Yeah, I'd anchor on the eval results. Novelty is mainly a scoring axis to help shape any refactoring, but evals should be the real driver of what content to include.

That said, I wonder if there is a way to indicate what information a skill _needs_ for efficacy, to shortcut review comments like this? 🤔

skills/mongodb-schema-design/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+
+### 2. Schema Fundamentals - 4 rules
+
+- [fundamental-embed-vs-reference](references/fundamental-embed-vs-reference.md) - Consult this reference for approaches to modeling different types of relationships (1:1, 1:few, 1:many, many:many, tree/hierarchical data) and how to decide between embedding and referencing based on access patterns.
+- [fundamental-document-model](references/fundamental-document-model.md) - Fundamentals of the document model. Consult this reference when migrating from SQL or other normalized data to a document database like MongoDB.
+- [fundamental-schema-validation](references/fundamental-schema-validation.md) - Consult this reference when creating new collections, or adding validation to existing collections, for example in response to finding inconsistent document structures or data quality issues.
+- [fundamental-document-size](references/fundamental-document-size.md) - Consult this reference when documents hit the hard 16MB limit, or when accesses are slower than expected as a result of large documents.
+
+### 3. Design Patterns - 11 rules
```
dacharyc, line 40: Thanks for your work fixing up the routing rules! Overall it seems much clearer when an agent should refer to any of these reference files, but I think this section could still use a little help. Here, we're still using terse descriptions like "Pre-calculate expensive aggregations" and "Collapse many optional fields into key-value attributes". These describe what the pattern does, not when an agent should load the file. An agent facing a schema problem can't determine which pattern files are relevant without reading them. Can we provide any more guidance here, similar to your updates above?

skills/mongodb-schema-design/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+| One-to-One | 1:1 | Always together | Embed |
+| One-to-Few | 1:N (N < 100) | Usually together | Embed array |
+| One-to-Many | 1:N (N > 100) | Often separate | Reference |
+| Many-to-Many | M:N | Varies | Two-way reference |
+
+This is a **rough** guideline, and whether to embed or reference depends on your specific access patterns, data size, and read/write frequencies. Always verify with your actual workload.
+
+## How to Use
+
+Read individual reference files for detailed explanations and code examples.
```
dacharyc: One nit: this could still be considered a directive/imperative instruction to "read all reference files" open to interpretation. Can we reword this to something like:

> Each reference file listed above contains detailed explanations and code examples. Use the descriptions in the Quick Reference to identify which files are relevant to your current task.

skills/mongodb-schema-design/references/fundamental-schema-validation.md
dacharyc: You've done really good work here to consolidate and trim! Just wanted to flag this new consideration I've observed after the refactoring:

Schema validation examples now appear in 3 files. The consolidation eliminated the old validation file overlap but introduced a new one. `$jsonSchema` validation examples now appear in:
  - fundamental-schema-validation.md (the primary reference)
  - fundamental-document-size.md:117–127 (array limits as prevention strategy)
  - fundamental-embed-vs-reference.md:119–138 (bounds enforcement for one-to-few)

Each serves a different purpose contextually, but an agent reading multiple files may see this as contradictory or redundant guidance. I'm wondering if a brief cross-reference note (e.g., "see fundamental-schema-validation for full validation guidance") would help scope-gate these. Did you observe any issues related to this in your testing?

skills/mongodb-schema-design/references/pattern-schema-versioning.md
```diff
@@ -0,0 +1,267 @@
...
+**Schema changes are inevitable, but uncontrolled changes cause schema drift** — documents in the same collection with inconsistent structures, leading to application errors and query failures. Use `schemaVersion` fields for safe migration and schema validation to prevent unexpected drift.
+
+### The problem: schema drift
+
+MongoDB's flexibility is a feature, but undisciplined field additions lead to code that must handle many document shapes.
+
+**Incorrect (uncontrolled drift over time):**
+
+```javascript
+// Over time, different versions of "user" documents accumulate
```
dacharyc, line 19: The schema drift examples (lines 19–33) show name/email field evolution, but the migration code (lines 88–134) operates on an address field that doesn't appear in these drift examples. An agent trying to follow the narrative as a coherent walkthrough could potentially find the switch from email-focused drift to address-focused migration confusing. Could we include an address variant here, or change the migration example to transform the email/name fields shown here?

skills/mongodb-schema-design/references/fundamental-document-size.md
```diff
@@ -0,0 +1,225 @@
...
+impact: CRITICAL
+impactDescription: "Hard 16MB BSON limit; oversized documents fail writes and degrade performance long before that"
+tags: schema, fundamentals, document-size, 16mb, bson-limit, arrays, anti-pattern, performance, indexing, subset-pattern, working-set, hot-data, cold-data, atlas-suggestion
+---
+
+## Keep Documents Small
+
+**MongoDB documents cannot exceed 16 megabytes.** This is a hard BSON limit, not a guideline — writes fail once a document reaches it.
+
+However, practical documents should be **much smaller than 16MB**. As a rule of thumb, aim for documents **under 1MB**. Smaller documents mean:
```
dacharyc, line 12: The 1MB suggestion here is a reasonable practical guideline, but it's presented with the same weight as the hard 16MB limit above it. Worth flagging whether this should be attributed (e.g., "MongoDB best practices recommend...") or softened further, since an agent may cite this as official MongoDB guidance.

skills/mongodb-schema-design/references/fundamental-embed-vs-reference.md
```diff
@@ -0,0 +1,416 @@
...
+// Publisher stays small and fixed-size
+{ _id: "oreilly", name: "O'Reilly Media", founded: 1978, bookCount: 3500 }
+
+// Each book references publisher; index on { publisherId: 1 }
+{ _id: "book001", title: "New MongoDB Book", publisherId: "oreilly" }
+
+// Efficient indexed queries
+db.books.find({ publisherId: "oreilly" })
+
+// $lookup when you need details from both sides
```
dacharyc, line 169: The earlier review from Asya said to use localField/foreignField syntax instead of `$expr` for join equality. Here, the one-to-many `$lookup` correctly uses localField/foreignField. However, the bidirectional many-to-many verification query at lines 386–398 uses `$expr` inside a `$lookup` pipeline. Wanted to check whether this is intentional (the pipeline form is needed for the double-match) or should be flagged for consistency with Asya's feedback.

dacharyc (commented):
I really appreciate your work on this, @addaleax  - this is shaping up nicely! Just a few more comments and Qs for consideration and then I think this will be good to go from my perspective.

skills/mongodb-schema-design/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+**⚠️ Security**: Use `--readOnly` for safety. Remove only if you need write operations.
+
+When connected, I can automatically:
+- Infer schema via `mcp__mongodb__collection-schema`
+- Measure document/array sizes via `mcp__mongodb__aggregate`
+- Check collection statistics via `mcp__mongodb__db-stats`
+
+### ⚠️ Action Policy
+
+**I will NEVER execute write operations without your explicit approval.**
```
doug-rodgers-mongodb, line 114: You might want to use something like the following to be a bit more explicit:

`Before any write via MCP, I will: (1) summarize the exact operation (collection, index/validator, estimated number of docs affected), and (2) ask for explicit confirmation (yes/no). I will not proceed on partial or ambiguous approvals.`

addaleax: Added!

skills/mongodb-schema-design/references/fundamental-embed-vs-reference.md
addaleax: Hm, that's a good point, I would defer to Asya but this can definitely be re-phrased in terms of localField/foreignField. However, part of the point of this snippet is to demonstrate the symmetry in the "match" checks, and this isn't a query that is intended to be made part of a production application, so I'm inclined to leave this as-is

skills/mongodb-schema-design/references/fundamental-document-size.md
addaleax: > The 1MB suggestion here is a reasonable practical guideline, but it's presented with the same weight as the hard 16MB limit above it.

This is intentional. I don't know if we have official guidance around document sizes, esp. given that the flexibility on that side is part of our selling point here, but 1MB is something I'd consider a genuine hard limit for design – those are usually still too large and will pollute data caches/slow down document loading beyond what would be optimal.

skills/mongodb-schema-design/references/pattern-schema-versioning.md
addaleax: Yup, switched to the email/name example the entire way 👍

skills/mongodb-schema-design/SKILL.md
addaleax: I've made some adjustments based on testing in 9219cec34430212b615dc538460f1a1a3fdf17fc, but honestly these seem to work quite well, and that's in line with my expectations. The examples you mentioned are actually pretty good – "Pre-calculate expensive aggregations" is to be used when dealing with expensive aggregations and "Collapse many optional fields into key-value attributes" with many optional fields. Like, these descriptions are already quite accurate, and while we could obviously add text here, if "short and sweet" does the job quite well, then anything else would just be increasing the token count, right?

skills/mongodb-schema-design/SKILL.md
addaleax: I don't know if there is a way to achieve that – I'm guessing comments are an option, but those count towards the token count and are processed by the LLM, right?

skills/mongodb-schema-design/references/fundamental-schema-validation.md
addaleax: I didn't observe any particular issues with this, no. But I also wouldn't expect any harm from the note you're suggesting, and it does make sense that the LLM could benefit from knowing about the fact that there are potential nuances around validation if it previously hadn't considered validation to be appicable, so I'll add it 👍

skills/mongodb-schema-design/SKILL.md
dacharyc: Alrighty, if you have found it's routing effectively in practice, it may not be necessary to be more explicit. My read has mainly been from the perspective of "if natural language is imprecise and can be interpreted differently across different runs, where can we add precision through explicitness and clarity." But I'm willing to lean on eval results where suggestions are speculative.

Thank you for testing!

skills/mongodb-schema-design/SKILL.md
dacharyc: Yeah, I'm wondering whether preemptive comments on PRs (not inline in the content itself) are a workaround for now. That way, a reviewer who is about to leave a comment on a section that seems like it may be redundant can see a comment already there from the PR author stating that the section improved the output with testing. But that puts extra burden on the PR author.

Separately, my delightfully nerdy wife wrote a system for private comments years ago, which can be shared across a team but not necessarily committed to the git history for a given project, and I'm wondering whether this is a good case to add that functionality for this project. Very speculative. Just thinking long-term about what good process might look like for creating new skills post-MVP.

skills/mongodb-schema-design/references/fundamental-embed-vs-reference.md
dacharyc: > this isn't a query that is intended to be made part of a production application

Oh! I hadn't thought about that lens. Have we made this distinction clear anywhere throughout this skill where we're showing queries? I'm wondering if there is a case we need to consider where someone uses this skill for schema design, and in the same session, writes queries to retrieve data based on the designed schema. Theoretically, with these queries already in context, an agent could apply any query patterns we show here instead of or in addition to anything already in it's base model training data or anything we're showing in the query-related skills.

I'm wondering if this would be a good thing to test for so we can validate we're not accidentally making query capabilities worse. 🤔

dacharyc (approved):
Thanks a ton for all your work on this, @addaleax . From my perspective, this looks to be in good shape and I'm happy for us to 🚢  assuming it clears SME review.

I did add a comment regarding a possible consideration to look into through testing, but it's realistically an edge case and is more of an open question, so I don't think it should block merging.

skills/mongodb-schema-design/SKILL.md
addaleax: Honestly, this PR here has received enough comments (including from Copilot) that it doesn't seem particularly helpful as reference material, simply because it's too noisy/long

I kind of would expect some semi-standardized solution to develop in the long-term for skills, this doesn't seem to something that only we would run into over time

(One other idea: We could also just add comments and have tooling that strips out the comments before they "ship". But I don't know if that's more or less practical than other approaches)

skills/mongodb-schema-design/references/fundamental-embed-vs-reference.md
addaleax: Yeah, this is similar to what we were talking about a week or so ago when we were discussing re-running eval cases once the initial versions of all skills are in, right? We want to make sure they don't interfere with each other.

I honestly haven't looked too much at the NLQ skill in #2, but it seems like that would be the right place to say things like "when using $lookup, prefer using localField and foreignField over a $match stage if possible"

Also, no, there hasn't been any distinction between "query/command to be run as part of the data modeling work itself" and "query/command that a user would run as part of their application". In the medium term, we'll want to make sure that the former category can be run through the MCP server, and in that case we may also want to translate them to MCP tool calls instead of using shell syntax as we currently do here. Currently, not all commands that we use here (e.g. setting the profiling level) can be run through the MCP server though. That's something @paula-stacho may want to look at for future iterations, but I don't think it needs to be part of the MVP.

PR #9, opened 2026-03-09
Add Search and AI Recommendations skill MCP-423

AyaanH123:


skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+```javascript
+db.collection.aggregate([
+  {
+    $vectorSearch: {
+      queryVector: [...],
+      numCandidates: 150,
+      limit: 50  // Get more candidates
+    }
+  },
+  { $limit: 20 },  // But only keep top 20 for merging
```
awjian: $vectorSearch.limit: 50 and $limit: 20 is redundant
```suggestion
      numCandidates: 150, // Get more candidates
      limit: 20  
    }
  }
```

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+  { $addFields: { score: { $meta: "vectorSearchScore" } } },
+  {
+    $unionWith: {
+      coll: "collection",
+      pipeline: [
+        { $search: { /* ... */ } },
+        { $limit: 20 }  // Also limit lexical results
+      ]
```
awjian: not sure how comprehensive you want to be but you need to have the same $addFields for the searchScore here

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+```javascript
+db.collection.aggregate([
+  {
+    $vectorSearch: {
+      queryVector: [...],
+      numCandidates: 150,
+      limit: 20
+    }
+  },
+  { $limit: 10 },
```
awjian: this should be after $rankFusion

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+    }
+  },
+  {
+    $group: {
+      _id: "$_id",
+      maxScore: { $max: "$score" },
+      doc: { $first: "$$ROOT" }
+    }
+  }
+])
```
awjian: this example doesn't use $scoreFusion?

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+- [Caching Strategies](#caching-strategies)
+- [Quick Reference](#quick-reference)
+
+---
+
+## Hybrid Search Optimizations
+
+### Limiting Results Between Stages
+
+In hybrid search, limit results from each pipeline before combining:
```
awjian: shouldn't you use $rankFusion for these instead of $unionWith?

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+
+---
+
+## Searching on Views
+
+Atlas Search indexes can be created on MongoDB views, which can be useful for specific scenarios.
+
+**When to use:**
+- Pre-filtered data: View contains only recent documents (e.g., movies from last 5 years)
+- Pre-joined data: View combines multiple collections that are frequently searched together
```
awjian: this is not supported

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+---
+
+## Searching on Views
+
+Atlas Search indexes can be created on MongoDB views, which can be useful for specific scenarios.
+
+**When to use:**
+- Pre-filtered data: View contains only recent documents (e.g., movies from last 5 years)
```
awjian:
```suggestion
- Pre-filtered "hot" data: View contains only recent documents (e.g., movies from last 5 years)
```

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+- Pre-joined data: View combines multiple collections that are frequently searched together
+- Transformed data: View performs expensive transformations once rather than at query time
+
+**Trade-offs:**
+- Views add overhead since the aggregation pipeline runs on each access
+- Generally slower than indexing the base collection directly
```
awjian: recommend using softer language here, e.g. "can slow down performance for complex view definitions"

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+- Index maintenance happens on view results, not raw data
+
+**Example use case:**
+```javascript
+// Create view of recent, highly-rated movies
+db.createView("recent_popular_movies", "movies", [
+  { $match: { year: { $gte: 2020 }, "imdb.rating": { $gte: 7 } } },
+  { $project: { title: 1, plot: 1, genres: 1, year: 1, "imdb.rating": 1 } }
+])
```
awjian: we only support $match stages that wrap a $expr

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+```javascript
+db.collection.explain("executionStats").aggregate([
+  { $search: { /* ... */ } }
+])
+```
+
+Look for:
+- `totalDocsExamined`: Lower is better
+- `executionTimeMillis`: Query execution time
+- `nReturned`: Number of results
```
awjian: would be useful to add this context to how search eplains differ from other mdb queries: https://wiki.corp.mongodb.com/spaces/MMS/pages/97457464/Runbook+Troubleshooting+Atlas+Search+Issues#Runbook:TroubleshootingAtlasSearchIssues-Question:HowtointerpretexplainplansonAtlasSearchqueries,andhowdotheydifferfromexplainplansonnon-$searchMongoDBqueries?


also more resources here for how to read search explain:
https://docs.google.com/presentation/d/1wDaUNFXlmngKb6zO2Vn2O7zlzPMb5O32JOMBaSev76g/edit?slide=id.g3489d9bff1d_0_49#slide=id.g3489d9bff1d_0_49

skills/search-and-ai-recommendations/references/optimization-general.md
```diff
@@ -0,0 +1,210 @@
...
+
+---
+
+## Search Nodes
+
+**Only recommend for very large workloads:**
```
awjian: this is not the guidance we want to offer, especially in light of recent incidents on coupled architecture

skills/search-and-ai-recommendations/references/optimization-search.md
```diff
@@ -0,0 +1,179 @@
...
+```javascript
+{
+  "mappings": {
+    "dynamic": true  // Index everything
+  }
+}
+```
+- **Pros**: Quick setup, works immediately
+- **Cons**: Larger index, slower queries, wastes resources on unused fields
+
```
awjian: need to add a new option for configurable dynamic indexing to specify 1) sub-documents to dynamically index and 2) the type of mapping to dynamically index

see "typeSets" in the docs

skills/search-and-ai-recommendations/references/optimization-search.md
```diff
@@ -0,0 +1,179 @@
...
+      "title": { "type": "string" },
+      "genre": { "type": "string", "analyzer": "lucene.keyword" }
+    }
+  }
+}
+```
+- **Pros**: Smaller index, faster queries, precise control
+- **Cons**: Requires knowing your schema
+
+**Recommendation:** Use `dynamic: false` in production for better performance.
```
awjian:
```suggestion
**Recommendation:** Use static mappings or a dynamic typeSet in production for optimized index size and performance.
```

skills/search-and-ai-recommendations/references/optimization-vector.md
```diff
@@ -0,0 +1,345 @@
...
+    "numDimensions": 1024,
+    "similarity": "cosine",
+    "quantization": {
+      "type": "binary"
+    }
+  }]
+}
+```
+
+**Decision guide:**
```
awjian: prakul has done a lot of thinking on this for auto-embedding, i recommend checking with him on these guidelines

skills/search-and-ai-recommendations/references/optimization-vector.md
```diff
@@ -0,0 +1,345 @@
...
+      queryVector: [...],
+      path: "embedding",
+      filter: {
+        text: {
+          query: "security",
+          path: "description"
+        }
+      }
+    }
+  }
```
awjian: this is incorrect, lexical prefilters is supported by $search, not $vectorSearch

skills/search-and-ai-recommendations/references/optimization-vector.md
```diff
@@ -0,0 +1,345 @@
...
+Voyage models support multiple output dimensions via `outputDimension` parameter:
+```javascript
+{
+  $vectorSearch: {
+    query: "text query",
+    path: "textField",
+    embeddingParameters: {
+      model: "voyage-4",
+      outputDimension: 512  // 256, 512, 1024, 2048, 4096
+    }
```
awjian: pretty sure this is hallucinated

awjian: embedding parameters are set at index time, not query time

skills/search-and-ai-recommendations/references/search-advanced-features.md
```diff
@@ -0,0 +1,357 @@
...
+- Field-specific: `fieldName:value`
+
+**Key considerations:**
+- Great for building search UIs with advanced options
+- Users can construct complex queries without API changes
+- Validate/sanitize user input to prevent injection
+
+---
+
+## Similar Items (moreLikeThis)
```
awjian: remove this, not frequently used operator

skills/search-and-ai-recommendations/references/search-advanced-features.md
```diff
@@ -0,0 +1,357 @@
...
+])
+```
+
+**Score aggregation options:**
+- `sum`: Add scores from all matching array elements
+- `maximum`: Use highest score from array elements
+- `minimum`: Use lowest score from array elements
+- `mean`: Average scores from array elements
+
+**Key considerations:**
```
awjian: performance can be degraded due to complexity of parent-child joins

awjian (changes requested):
did a first pass, few recommendations and corrections for accuracy

skills/search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
+---
+name: search-and-ai
+description: |
+  Guides MongoDB users through implementing and optimizing Atlas Search (full-text), Vector Search (semantic), and Hybrid Search solutions. Use this skill when users need to build search functionality, whether for text-based queries (autocomplete, fuzzy matching, faceted search), semantic similarity (embeddings, RAG applications), or combined approaches. Provides workflows for selecting the right search type, creating indexes, constructing queries, and optimizing performance using the MongoDB MCP server.
```
dacharyc: Nit: the scope gating here is general - "search functionality" and then specific - "text-based queries" etc. If we can drop the "whether" here and make it specifically these use cases, I think that could help trigger efficacy.

```suggestion
  Guides MongoDB users through implementing and optimizing Atlas Search (full-text), Vector Search (semantic), and Hybrid Search solutions. Use this skill when users need to build search functionality for text-based queries (autocomplete, fuzzy matching, faceted search), semantic similarity (embeddings, RAG applications), or combined approaches. Provides workflows for selecting the right search type, creating indexes, constructing queries, and optimizing performance using the MongoDB MCP server. 
```

skills/mongodb-search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+
+# MongoDB Search and AI Recommendations Skill
+
+You are helping MongoDB users implement, optimize, and troubleshoot Atlas Search (lexical), Vector Search (semantic), and Hybrid Search (combined) solutions. Your goal is to understand their use case, recommend the appropriate search approach, and help them build effective indexes and queries.
+
+## Core Principles
+
+1. **Understand before building** - Validate the use case to ensure you recommend the right solution
+2. **Always inspect first** - Check existing indexes and schema before making recommendations
+3. **Explain before executing** - Describe what indexes will be created and require explicit approval
```
dacharyc, line 15: This is an excellent scope gate - thank you for including it! 🎉

skills/search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+Once you have determined the search type, always consult the appropriate reference file(s) before recommending indexes or queries:
+- **Lexical**: consult both `references/lexical-search-indexing.md` (index) and `references/lexical-search-querying.md` (query)
+- **Vector**: consult `references/vector-search.md`
+- **Hybrid**: consult `references/hybrid-search.md` (and the lexical/vector files for the individual pipeline stages within it)
+
+### 3. Query Construction
+
+Consult the appropriate reference file for full query syntax and examples. When constructing queries, keep in mind:
+
+**Compound query clauses (Atlas Search):**
```
dacharyc: Since this is only about Atlas Search, it probably makes sense to move this to the `lexical-search-querying.md` (or maybe duplicating it there and in `hybrid-search.md` if it is also relevant to hybrid search?)

In other words, if it's not needed for vector search, we could remove it from the SKILL.md and avoid potentially confusing agents with these clauses when they load this into context and then go through to the Vector Search path.

skills/search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+- `must`: Required matches that affect scoring
+- `should`: Optional matches that boost scores
+- `filter`: Required matches that don't affect scoring (faster)
+- `mustNot`: Exclusions
+
+### 5. Optimization Considerations
+
+Before finalizing recommendations, review:
```
dacharyc: It seems like we're mixing together different types of query paths with these fields. Can we move optimization considerations into the relevant search type where they're relevant?

skills/search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+
+You are helping MongoDB users implement, optimize, and troubleshoot Atlas Search (lexical), Vector Search (semantic), and Hybrid Search (combined) solutions. Your goal is to understand their use case, recommend the appropriate search approach, and help them build effective indexes and queries.
+
+## Core Principles
+
+1. **Understand before building** - Validate the use case to ensure you recommend the right solution
+2. **Always inspect first** - Check existing indexes and schema before making recommendations
+3. **Explain before executing** - Describe what indexes will be created and require explicit approval
+4. **Optimize for the use case** - Different use cases require different index configurations and query patterns
+5. **Handle read-only scenarios** - If users have --readOnly flag set, explain optimal configurations without creating indexes
```
dacharyc: Do we need to explain to agents how to find the value of this flag - i.e. will the agent know if the user has `--readOnly` flag set?

Just looked around for info about this and found this:

> When read-only mode is enabled, the server will not register tools that have "create", "update", or "delete" operation types, effectively preventing AI clients or users from performing write operations. The default behavior of the MongoDB MCP Server is to allow cluster write operations, so you must explicitly enable this flag for read-only access.

So for an agent, it might look like:

> If you do not have access to 'create', 'update', or 'delete' operation tools, you are in read-only mode. Apply alternative workflows labeled 'read-only mode' during this session.

skills/mongodb-search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+- **storedSource**: Store frequently projected/filtered fields in the index to avoid full document lookups
+- **exact: true**: For small collections (<10K docs) or accuracy-critical queries, use exact nearest neighbor
+
+### 6. Execution and Validation
+
+**Creating indexes:**
+1. Explain the index configuration in plain language
+2. Show the JSON structure
+3. Ask what the user wants to name the index
+4. Get explicit approval: "Should I create this index?"
```
dacharyc, line 93: Yes, excellent scope gating here - thank you! 🎉

skills/search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+5. Use MCP's `create-index` tool after approval
+6. In read-only mode, provide the complete index JSON for creation via the Atlas UI
+
+**Running queries:**
+1. Show the aggregation pipeline
+2. Execute using MCP's `aggregate` tool
+3. Present results clearly
+4. Use `explain` to analyze query performance if needed
+
+**Refining existing queries/indexes:**
```
dacharyc: The "Refining" workflow here doesn't connect these details to what an agent actually needs to do to complete these steps. The four steps are entirely abstract; "identify issues," "propose improvements," "test and compare"; with no directive telling the agent to consult the optimization sections of the relevant reference file. Step 2 *does* contain this info - (Determine Search Type), which explicitly says "consult references/vector-search.md" etc. - but this path has no equivalent routing.

I suggest something like adding a single line connecting the dots, plus making the abstract steps concrete:

```
1. Review current implementation (inspect indexes with `collection-indexes`, ask user to share query code)
2. Compare against the optimization and best practices guidance in the reference file(s) consulted in Step 2
3. Propose specific improvements with before/after examples
4. Run the revised query with `aggregate` and use `explain` to compare performance
```

skills/mongodb-search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+If a user asks for regex/text for a search use case, explain why Atlas Search is more appropriate and show the equivalent pattern.
+
+## Reference Files
+
+- `references/lexical-search-indexing.md` — Creating or modifying an Atlas Search index: field types, analyzers, dynamic vs explicit vs typeSet mappings, storedSource configuration, synonym mapping setup, creating Search indexes on Views
+- `references/lexical-search-querying.md` — Writing a $search or $searchMeta query: $search vs $searchMeta, compound operator, queryString, embeddedDocument, highlighting, returnStoredSource at query time, multi-analyzer path syntax, explain
+- `references/vector-search.md` — Creating a Vector Search index or writing a $vectorSearch query: index definition, filter fields, numCandidates tuning, ANN vs ENN, pre/post filtering, creating Vector Search indexes on Views
+- `references/hybrid-search.md` — Combining search methods: $rankFusion, $scoreFusion, lexical prefilters using the vectorSearch operator inside $search, required index setup for hybrid, choosing between $rankFusion/$scoreFusion/lexical prefilter
+
+## Handling Edge Cases
```
dacharyc, line 116: This section mixes concerns across all three search paradigms - can we move out the relevant details to the relevant files? We don't need to use the tokens loading this into context for all search types, and having inapplicable instructions for a specific type in context could potentially confuse an agent.

skills/search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+
+Consult the appropriate reference file for full query syntax and examples. When constructing queries, keep in mind:
+
+**Compound query clauses (Atlas Search):**
+- `must`: Required matches that affect scoring
+- `should`: Optional matches that boost scores
+- `filter`: Required matches that don't affect scoring (faster)
+- `mustNot`: Exclusions
+
+### 5. Optimization Considerations
```
dacharyc: We're jumping from 3 to 5 here - can we renumber this and the next section?

skills/mongodb-search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Vector Search (Semantic):**
+Use when users need:
+- Semantic similarity ("find movies about coming of age stories")
+- Natural language understanding
+- RAG (Retrieval Augmented Generation) applications
+- Finding conceptually similar items
+- Cross-modal search
+
+**Hybrid Search:**
```
dacharyc, line 63: This is outside of my area of expertise, but in reviewing the reference files, it looks like there are several version-specific references where new functionality has been introduced. Claude suggests these are concentrated in hybrid search, and affect whether hybrid search is available. If that's accurate, we should scope gate this path - if users are on older versions where relevant functionality isn't available, we should not ask agents to spend time trying to implement it. Instead, it should let the user know the functionality isn't available, and if possible, offer an alternative or suggest upgrading.

dacharyc (commented):
Hey @AyaanH123 - overall this looks good, but I have a handful of suggestions for SKILL.md to move content out of the main skill body where it is only applicable to one of the search types, and a few similar optimizations we might make.

I'll follow up with specific suggestions related to the references.

skills/mongodb-search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+
+**Example:**
+If user wants "car" to also find "vehicle" and "automobile":
+1. Tell user to create collection: db.synonyms.insertOne({ "mappingType": "equivalent", "synonyms": ["car", "vehicle", "automobile"] })
+2. Add to index with analyzer matching the field being searched
+3. Queries automatically expand (user searches "car", MongoDB Search searches: car OR vehicle OR automobile)
+
+---
+
+## Searching on Views
```
dacharyc, line 547: lexical-search-indexing.md (lines 548-586) and vector-search.md (lines 744-799) both contain "Searching on Views" sections with the same:
  - Version requirements (8.0+, 8.1+ for programmatic creation)
  - Supported view stages ($addFields, $set, $match with $expr only)
  - Troubleshooting (FAILED, STALE states)
  - The same 8.0 vs 8.1 query behavior explanation

The main difference is the examples (one uses $search, the other uses $vectorSearch). This is ~35-40 lines of duplicated content. If a hybrid search user loads both files, they get this content twice. We may want to consider consolidating into one file and cross-referencing from the other, or extracting a shared views.md reference.

skills/mongodb-search-and-ai/references/hybrid-search.md
```diff
@@ -0,0 +1,712 @@
...
+          "similarity": "dotProduct",
+          "quantization": "scalar"  // Optional
+        }
+      }
+    }
+  }
+)
+```
+
+**Note**: `storedSource: true` is not supported on indexes that contain a `vector` field type. Use `include` or `exclude` to specify stored fields explicitly.
```
dacharyc, line 111: The rule "storedSource is not supported on indexes with vector field type" appears at:
  - Line 96 (indexing section)
  - Line 155 ($rankFusion restrictions)
  - Line 364 ($scoreFusion restrictions)
  - Line 710 (best practices/limitations)

Similarly, "input pipelines cannot contain $project or storedSource fields" appears at lines 155, 364, and 693. An agent reading this file encounters the same constraint 3-4 times. We could consolidate if we state it once authoritatively (e.g., in the indexing section or a constraints preamble) and don't repeat it.

skills/search-and-ai/references/hybrid-search.md
```diff
@@ -0,0 +1,712 @@
...
+    }
+  }
+)
+```
+
+**Note**: `storedSource: true` is not supported on indexes that contain a `vector` field type. Use `include` or `exclude` to specify stored fields explicitly.
+
+---
+
+## $rankFusion
```
dacharyc: It looks like `$rankFusion` and `$scoreFusion` share boilerplate in this file.

These two sections have significant structural overlap:
  - Both explain pipeline naming restrictions (line 149 — stated once but applies to both)
  - Both list input pipeline restrictions with nearly identical language (lines 153-157 vs 362-366)
  - Both explain `scoreDetails` surfacing with the same pattern

Consider a shared "Common rules for fusion stages" section before the two operator-specific sections, covering naming restrictions, pipeline restrictions, and scoreDetails. Then each operator section only covers what's unique to it (RRF formula vs normalization options, weights vs expressions).

skills/mongodb-search-and-ai/references/vector-search.md
```diff
@@ -0,0 +1,800 @@
...
+])
+```
+
+**Performance Note**: Post-filtering is slower than pre-filtering because it computes similarity for all candidates first.
+
+---
+
+## Query Optimization
+
+### numCandidates Tuning
```
dacharyc, line 584: This section may be more verbose than needed.

The numCandidates guidance spans lines 584-664 — roughly 80 lines covering:
  - Rule of thumb (20x limit)
  - Three example code blocks showing 15x, 50x, and 5x ratios
  - "Factors Affecting numCandidates Selection" subsection
  - "When to Adjust numCandidates" subsection with increase/decrease lists

The three code blocks differ only in the numCandidates value and a comment. One example with a note about the range (10-50x depending on requirements) would convey the same information. The "Factors" and "When to Adjust" sections also overlap - "Quantized vectors reduce accuracy" appears as a factor, and "Using quantized vectors" appears as a reason to increase. This could be condensed to roughly half its current size. We can probably improve token efficiency here, and for the hybrid case that loads all files, with these changes.

skills/mongodb-search-and-ai/references/hybrid-search.md
dacharyc, line 1: We're missing a version-requirement preamble here.

The best scope gate would be a mention in SKILL.md version scope-gating for hybrid search. But if an agent makes it to this file, version requirements are scattered across individual sections (line 27 table, line 102 for $rankFusion, line 311 for $scoreFusion, lines 700-702 in limitations). An agent entering this file doesn't get an upfront "this file requires MongoDB 8.0+ minimum" gate. If the user is on 7.x, the agent could read through substantial content before discovering it doesn't apply. A brief version prerequisite at the top would help. And then specific version info at any section that requires it, so agents can pay less attention to the info if the required version is not available.

skills/mongodb-search-and-ai/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+2. **Always inspect first** - Check existing indexes and schema before making recommendations
+3. **Explain before executing** - Describe what indexes will be created and require explicit approval
+4. **Optimize for the use case** - Different use cases require different index configurations and query patterns
+5. **Handle read-only scenarios** - If users have --readOnly flag set, explain optimal configurations without creating indexes
+
+## Workflow
+
+### 1. Discovery Phase
+
+**Check the environment:**
```
dacharyc, line 23: Given some of the things in hybrid search require specific MongoDB versions, we should probably add some sort of version check to this environment check so the agent knows what functionality is available.

skills/search-and-ai/references/hybrid-search.md
```diff
@@ -0,0 +1,712 @@
...
+
+Hybrid search combines multiple search methods on the same collection and merges the results into a single ranked or scored list.
+
+**Three patterns covered in this guide:**
+
+| Pattern | Stage / Operator | MongoDB Version | Use When |
+|---|---|---|---|
+| Rank-based fusion | `$rankFusion` | 8.0+ | Document position matters; use RRF algorithm |
+| Score-based fusion | `$scoreFusion` | 8.2+ | Score magnitude matters; need custom math or normalization |
+| Lexical prefilter | `$search` + `vectorSearch` operator | Preview | Need fuzzy/phrase/wildcard/compound pre-filtering before vector search |
```
dacharyc: We mention here and down below that this is "Preview" - but not what the implications are for the agent. We should explicitly provide details for the agent about the implications of this being a preview. i.e.: "APIs may change" "current implementation is not guaranteed to match production behavior when this is promoted to production" (or even that it may not be promoted to production?) - whatever we mean at MongoDB when we call something "Preview." If an agent is implementing search in a production environment that requires stable functionality, preview functionality may not be suitable.

skills/search-and-ai/references/lexical-search-querying.md
```diff
@@ -0,0 +1,383 @@
...
+
+`$searchMeta` shares the following fields with `$search`: `index`, all operator names (e.g. `text`, `range`, `compound`), `concurrent`, and `returnStoredSource`.
+
+`$searchMeta`-only fields:
+
+| Field | Description |
+|---|---|
+| `count` | Returns total or lower-bound count of matching documents. `{ "type": "total \| lowerBound" }` |
+| `facet` | Collector that returns facet bucket metadata instead of an operator |
+| `returnScope` | Sets query context to an embedded document field. Requires `returnStoredSource: true` on MongoDB < 8.2 |
```
dacharyc: This is mentioned in this table but never elaborated. I didn't see any example, or explanation of when to use it, and no connection to any workflow in SKILL.md. An agent encountering this would have no idea what to do with it. We might want to add a brief usage example or remove it if it's outside the skill's scope.

dacharyc (commented):
Handful more comments across the reference files, which I didn't get to in the initial review.

skills/search-and-ai/SKILL.md
AyaanH123: Updated

skills/search-and-ai/SKILL.md
AyaanH123: Updated verbiage to mention that instead of trying to build the index, to provide the configuration + allude to the UI

skills/mongodb-search-and-ai/SKILL.md
AyaanH123: Agreed - will specifically call out the atlas-inspect-cluster tool and leverage the result later on

skills/mongodb-search-and-ai/SKILL.md
AyaanH123: added a step to ensure that we don't reference the hybrid search file unnecessarily if the user's cluster version doesn't support the intended use case

skills/search-and-ai/SKILL.md
AyaanH123: Agreed this is not needed here and is covered in more depth in the `lexical-search-querying.md` - removing

skills/mongodb-search-and-ai/SKILL.md
AyaanH123: Yup, removing anything repetitive and search-type-specific to respective files

skills/mongodb-search-and-ai/references/lexical-search-indexing.md
AyaanH123: will take the reference approach

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+```javascript
+// Field-level (add to mappings.fields):
+{
+  "title": {
+    "type": "string",
+    "analyzer": "lucene.standard"  // Or omit to use default
+  },
+  "category": {
+    "type": "string",
+    "analyzer": "lucene.keyword"   // Exact matching
```
awjian: for exact matching, recommend using `token` instead of the string type

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+**Dynamic mapping includes:** boolean, date, number, objectId, string, uuid
+**Must configure explicitly:** autocomplete, token, geo, embeddedDocuments, vector
+
+### Quick Reference
+
+| Type | When to Use | Required Fields | Optional Fields (with valid values) | Notes |
+|------|-------------|-----------------|-------------------------------------|-------|
+| **string** | Full-text search, phrase matching, fuzzy search | type | analyzer, searchAnalyzer, indexOptions: "docs" or "freqs" or "positions" or "offsets", store: true or false, multi | Default for text. For sorting use token instead. |
+| **token** | Sort/facet on text, exact matching | type | normalizer: "lowercase" or "none" (default: "none") | Required for sorting strings. Max 8181 chars. |
+| **autocomplete** | Search-as-you-type, typeahead | type | analyzer, tokenization: "edgeGram" or "rightEdgeGram" or "nGram" (default: "edgeGram"), minGrams (default: 2), maxGrams (default: 15), foldDiacritics: true or false | Not in dynamic mapping. Recommend maxGrams ≤ 15. |
```
awjian:
```suggestion
| **autocomplete** | Search-as-you-type, typeahead, partial or substring matching | type | analyzer, tokenization: "edgeGram" or "rightEdgeGram" or "nGram" (default: "edgeGram"), minGrams (default: 2), maxGrams (default: 15), foldDiacritics: true or false | Not included in "dynamic: true". Recommend maxGrams ≤ 15. |
```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+| **boolean** | True/false filters | type | None | Included in dynamic mapping. |
+| **date** | Date ranges, timestamps | type | None | Use range operator for arrays. |
```
awjian: not really sure what the range operator is relevant for here
```suggestion
| **date** | Date ranges, timestamps | type | None | Included in "dynamic: true" |
```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+**Must configure explicitly:** autocomplete, token, geo, embeddedDocuments, vector
+
+### Quick Reference
+
+| Type | When to Use | Required Fields | Optional Fields (with valid values) | Notes |
+|------|-------------|-----------------|-------------------------------------|-------|
+| **string** | Full-text search, phrase matching, fuzzy search | type | analyzer, searchAnalyzer, indexOptions: "docs" or "freqs" or "positions" or "offsets", store: true or false, multi | Default for text. For sorting use token instead. |
+| **token** | Sort/facet on text, exact matching | type | normalizer: "lowercase" or "none" (default: "none") | Required for sorting strings. Max 8181 chars. |
+| **autocomplete** | Search-as-you-type, typeahead | type | analyzer, tokenization: "edgeGram" or "rightEdgeGram" or "nGram" (default: "edgeGram"), minGrams (default: 2), maxGrams (default: 15), foldDiacritics: true or false | Not in dynamic mapping. Recommend maxGrams ≤ 15. |
+| **boolean** | True/false filters | type | None | Included in dynamic mapping. |
```
awjian: more specific 
```suggestion
| **boolean** | True/false filters | type | None | Included in "dynamic: true" |
```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+| **date** | Date ranges, timestamps | type | None | Use range operator for arrays. |
+| **number** | Numeric queries, ranges, sorting | type | representation: "int64" or "double" (default: "double"), indexIntegers: true or false, indexDoubles: true or false | Use int64 for large integers. |
+| **objectId** | Query by _id | type | None | Standard MongoDB ObjectIds. |
+| **uuid** | UUID identifiers | type | None | BSON Binary Subtype 4. |
+| **geo** | Location search, geographic queries | type | indexShapes: true or false (default: false) | Requires GeoJSON. Set indexShapes=true for polygons. |
+| **embeddedDocuments** | Search in arrays of objects, independent scoring | type | dynamic: true or false or {typeSet: "name"} (default: false), fields, storedSource: true or false or {include/exclude} | Not in dynamic mapping. Max 5 nesting levels. Each doc counts toward 2.1B limit. |
```
awjian:
```suggestion
| **embeddedDocuments** | Search in arrays of objects, independent scoring | type | dynamic: true or false or {typeSet: "name"} (default: false), fields, storedSource: true or false or {include/exclude} | Not included in "dynamic: true". Max 5 nesting levels. Each nested document counts toward 2.1B limit. |
```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+| **token** | Sort/facet on text, exact matching | type | normalizer: "lowercase" or "none" (default: "none") | Required for sorting strings. Max 8181 chars. |
+| **autocomplete** | Search-as-you-type, typeahead | type | analyzer, tokenization: "edgeGram" or "rightEdgeGram" or "nGram" (default: "edgeGram"), minGrams (default: 2), maxGrams (default: 15), foldDiacritics: true or false | Not in dynamic mapping. Recommend maxGrams ≤ 15. |
+| **boolean** | True/false filters | type | None | Included in dynamic mapping. |
+| **date** | Date ranges, timestamps | type | None | Use range operator for arrays. |
+| **number** | Numeric queries, ranges, sorting | type | representation: "int64" or "double" (default: "double"), indexIntegers: true or false, indexDoubles: true or false | Use int64 for large integers. |
+| **objectId** | Query by _id | type | None | Standard MongoDB ObjectIds. |
+| **uuid** | UUID identifiers | type | None | BSON Binary Subtype 4. |
+| **geo** | Location search, geographic queries | type | indexShapes: true or false (default: false) | Requires GeoJSON. Set indexShapes=true for polygons. |
```
awjian:
```suggestion
| **number** | Numeric queries, ranges, sorting | type | representation: "int64" or "double" (default: "double"), indexIntegers: true or false, indexDoubles: true or false | Included in "dynamic: true" | Use int64 for large integers. |
| **objectId** | Query by _id | type | None | Included in "dynamic: true" | Standard MongoDB ObjectIds. |
| **uuid** | UUID identifiers | type | None | Included in "dynamic: true" | BSON Binary Subtype 4. |
| **geo** | Location search, geographic queries | type | indexShapes: true or false (default: false) | Requires GeoJSON. Set indexShapes=true for polygons. | Included in "dynamic: true".
```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+| **embeddedDocuments** | Search in arrays of objects, independent scoring | type | dynamic: true or false or {typeSet: "name"} (default: false), fields, storedSource: true or false or {include/exclude} | Not in dynamic mapping. Max 5 nesting levels. Each doc counts toward 2.1B limit. |
+| **vector** | Lexical prefilters for semantic search | type, numDimensions (1-8192), similarity: "cosine" or "dotProduct" or "euclidean" | quantization: "none" or "scalar" or "binary" (default: "none"), hnswOptions.maxEdges: 16-64, hnswOptions.numEdgeCandidates: 100-3200 | For hybrid search. See vector-search.md and hybrid-search.md. |
```
awjian:
```suggestion
| **vector** | Lexical prefilters for semantic search | type, numDimensions (1-8192), similarity: "cosine" or "dotProduct" or "euclidean" | quantization: "none" or "scalar" or "binary" (default: "none"), hnswOptions.maxEdges: 16-64, hnswOptions.numEdgeCandidates: 100-3200 | For hybrid search. See vector-search.md and hybrid-search.md. | Not included in "dynamic: true".
```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+- User is prototyping or exploring data with unknown schema
+- Need to get started quickly without defining all fields
+- All or most fields need to be searchable
+- Accept larger index size and slower performance for convenience
+
+**Use explicit (dynamic: false)** when:
+- User knows exactly which fields to search
+- Performance and index size are priorities
+- Schema is stable and well-defined
+- Only a subset of fields need to be searchable
```
awjian:
```suggestion
**Use explicit (dynamic: false)** when:
- User has completed early stages of prototyping and knows exactly which fields to search
- Performance and index size are priorities
- Schema is stable and well-defined
- Only a subset of fields need to be searchable

```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+
+**Choose based on user's needs:**
+
+**Use dynamic (true)** when:
+- User is prototyping or exploring data with unknown schema
+- Need to get started quickly without defining all fields
+- All or most fields need to be searchable
+- Accept larger index size and slower performance for convenience
+
+**Use explicit (dynamic: false)** when:
```
awjian:
```suggestion
**Use explicit (dynamic: false)** (recommended for production) when:
```

skills/mongodb-search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+- Accept larger index size and slower performance for convenience
+
+**Use explicit (dynamic: false)** when:
+- User knows exactly which fields to search
+- Performance and index size are priorities
+- Schema is stable and well-defined
+- Only a subset of fields need to be searchable
+
+**Use typeSets (recommended for production)** when:
+- User wants automatic indexing but with control over which types
```
awjian, line 320:
```suggestion
- User wants automatic indexing but with control over which types
- Document schema is dynamic and new fields need to be indexed automatically without an index rebuild
```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+        }
+      ]
+    }
+  ]
+}
+```
+- **Pros**: Automatically indexes specified field types, more control than full dynamic, can configure different typeSets for sub-documents
+- **Cons**: Still indexes all fields of specified types
+
+**Recommendation:** Use static mappings or a dynamic typeSet in production for optimized index size and performance.
```
awjian:
```suggestion
**Recommendation:** Use static mappings or a dynamic typeSet (within a specific path, not at the root document level) in production for optimized index size and performance.
```

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -0,0 +1,585 @@
...
+**Examples:**
+
+**Store specific fields:**
+```javascript
+{
+  "mappings": {
+    "dynamic": false,
+    "fields": {
+      "title": { "type": "string" },
+      "genre": { "type": "string", "analyzer": "lucene.keyword" },
```
awjian:
```suggestion
      "genre": { "type": "token" },
```

skills/search-and-ai/references/lexical-search-querying.md
```diff
@@ -0,0 +1,580 @@
...
+Cursor-based pagination using tokens. More efficient than `$skip` alone for deep pagination.
+
+**Step 1 — Get tokens from the initial query:**
+```javascript
+db.movies.aggregate([
+  {
+    $search: {
+      index: "<index-name>",
+      text: { path: "title", query: "summer" },
+      sort: { released: 1 }  // Sort on a unique field to prevent tie-ordering issues
```
awjian:
```suggestion
      sort: { released: 1, _id: 1 }  // Sort on a unique field to prevent tie-ordering issues
```

skills/search-and-ai/references/lexical-search-querying.md
```diff
@@ -0,0 +1,580 @@
...
+
+**Step 2 — Next page using searchAfter:**
+```javascript
+db.movies.aggregate([
+  {
+    $search: {
+      index: "<index-name>",
+      text: { path: "title", query: "summer" },
+      searchAfter: "<token-from-last-document-on-previous-page>",
+      sort: { released: 1 }
```
awjian:
```suggestion
      sort: { released: 1, _id: 1 } // maintain the same sort order
```

skills/search-and-ai/references/lexical-search-querying.md
```diff
@@ -0,0 +1,580 @@
...
+      range: { path: "funding_rounds.raised_amount", gte: 5000000, lte: 10000000 },
+      returnStoredSource: true,
+      returnScope: { path: "funding_rounds" }
+    }
+  },
+  { $limit: 5 }
+])
+```
+
+Only fields defined in `storedSource` within the embedded document are returned — root-level fields are excluded.
```
awjian:
```suggestion
Only fields defined in `storedSource` within the embedded document are returned — root-level fields are excluded. When `returnScope` is specified, all query paths must start with `returnScope.path`
```

skills/search-and-ai/references/lexical-search-querying.md
```diff
@@ -0,0 +1,580 @@
...
+**Key considerations:**
+- Great for building search UIs with advanced options
+- Users can construct complex queries without API changes
+- Validate/sanitize user input to prevent injection
+
+---
+
+### Searching Nested Arrays (embeddedDocument)
+
+**Use case:** Search within arrays of objects where each element should be scored independently.
```
awjian:
```suggestion
**Use case:** Search within arrays of objects where each element-wise comparisons are required (similar to $elemMatch), or each element must be scored independently.
```

skills/mongodb-search-and-ai/references/lexical-search-querying.md
```diff
@@ -0,0 +1,580 @@
...
+Query geographic points within a region. Field must be indexed as `geo` type. Specify one of:
+- `box`: `{ bottomLeft: <GeoJSON Point>, topRight: <GeoJSON Point> }`
+- `circle`: `{ center: <GeoJSON Point>, radius: <meters> }`
+- `geometry`: GeoJSON Polygon or MultiPolygon
+
+**For both geo operators:** longitude must be specified before latitude; longitude range [-180, 180], latitude range [-90, 90].
+
+---
+
+## Query Optimization
```
awjian, line 543: was there sections on $sort and $match after $search here?

awjian (changes requested):

dacharyc (approved):
Thanks for incorporating my feedback, @AyaanH123 ! I've done another pass here and it looks like you've successfully addressed everything I raised, and nothing new has been introduced, so I'm feeling good about this from my side.

If you end up making more substantive changes as a result of awjian's pending feedback, I'll probably want to make one final pass across those files, but I don't anticipate anything blocking so I'll go ahead and ✅  here for expediency.

dacharyc (approved):
Looks good from my side - I think once @awjian has had a chance to review any outstanding technical comments, we're good to merge. ✅ 

One thing I would call out - it would be good to test this with the query optimization skill once both are merged. I see "optimizing" in the trigger description in a couple of places here, and I wonder if there may be any overlap or mis-triggering once both are in. Non-blocking, just something to check after merge.

skills/search-and-ai/SKILL.md
```diff
@@ -0,0 +1,139 @@
+---
+name: search-and-ai
```
dacharyc: I just noticed we have namespaced all of the other MongoDB skills, which is a good practice - we should do that here, and also update the directory name to match - i.e. `mongodb-search-and-ai`

```suggestion
name: mongodb-search-and-ai
```

dacharyc (commented):
One last comment!

skills/search-and-ai/references/lexical-search-indexing.md
```diff
@@ -265,15 +264,15 @@ Normalizers produce a single token (used with `token` field type):
 |------|-------------|-----------------|-------------------------------------|-------|
 | **string** | Full-text search, phrase matching, fuzzy search | type | analyzer, searchAnalyzer, indexOptions: "docs" or "freqs" or "positions" or "offsets", store: true or false, multi | Default for text. For sorting use token instead. |
 | **token** | Sort/facet on text, exact matching | type | normalizer: "lowercase" or "none" (default: "none") | Required for sorting strings. Max 8181 chars. |
```
awjian: "Required for sorting or faceting strings."

awjian (approved):
LGTM % one nit! Great job!

dacharyc (approved):
LGTM - thanks for making that name change! ✅

PR #10, opened 2026-03-09
Add Atlas Stream Processing skill MCP-422

laurazhukas:
### Overview
This PR adds a comprehensive skill for building, operating, and debugging MongoDB Atlas Stream Processing (ASP) pipelines through the MongoDB MCP Server.

### What's Included

**Main Skill** (\`SKILL.md\`)
- Complete workflow guidance for workspace, connection, and processor management
- Four MCP tools: \`atlas-streams-discover\`, \`atlas-streams-build\`, \`atlas-streams-manage\`, \`atlas-streams-teardown\`
- Pre-deployment validation checklists
- Billing and cost awareness guidance
- Comprehensive troubleshooting section

**Reference Documentation** (5 files)
- \`connection-configs.md\` - Connection schemas by type (Kafka, Cluster, S3, HTTPS, Lambda, etc.)
- \`development-workflow.md\` - 5-phase incremental development lifecycle
- \`output-diagnostics.md\` - Processor health classification and diagnostic framework
- \`pipeline-patterns.md\` - Source/sink patterns, windowing, enrichment examples
- \`sizing-and-parallelism.md\` - Tier selection, parallelism tuning, cost optimization

### Key Features

✅ **Comprehensive coverage**: Workspaces, connections, processors, networking  
✅ **Best practices**: Pipeline validation, connection verification, error handling  
✅ **Production-ready**: DLQ configuration, tier sizing, performance tuning  
✅ **Safety-first**: Billing warnings, destructive action confirmations  
✅ **Integration-focused**: Kafka, S3, Lambda, HTTPS endpoints  

### Skill Metadata
- **Name**: \`atlas-stream-processing\`
- **Version**: 1.0.0
- **User-invocable**: Yes

### Testing
This skill has been tested with MongoDB MCP Server and successfully manages:
- Multi-workspace environments
- Complex Kafka→Atlas and Atlas→Kafka pipelines
- Processor debugging and health diagnostics
- Connection configuration and validation

### DIFF
--- a/skills/atlas-stream-processing/SKILL.md
+++ b/skills/atlas-stream-processing/SKILL.md
@@ -14,7 +14,6 @@ Build, operate, and debug Atlas Stream Processing (ASP) pipelines using four MCP
 
 This skill requires the **MongoDB MCP Server** connected with:
 - Atlas API credentials (`apiClientId` and `apiClientSecret`)
-- `previewFeatures: ["streams"]` enabled in the MCP server config
 
 The 4 tools: `atlas-streams-discover`, `atlas-streams-build`, `atlas-streams-manage`, `atlas-streams-teardown`.
 
@@ -36,7 +35,6 @@ If the MongoDB MCP Server is not connected or the streams tools are missing, see
 | `list-processors` | See all processors in a workspace |
 | `inspect-processor` | Check processor state, pipeline, config |
 | `diagnose-processor` | Full health report: state, stats, errors |
-| `get-logs` | Operational logs (runtime errors) or audit logs (lifecycle) |
 | `get-networking` | PrivateLink and VPC peering details |
 
 **Pagination** (all list actions): `limit` (1-100, default 20), `pageNum` (default 1).
@@ -48,14 +46,14 @@ If the MongoDB MCP Server is not connected or the streams tools are missing, see
 | `workspace` | `cloudProvider`, `region`, `tier` (default SP10), `includeSampleData` |
 | `connection` | `connectionName`, `connectionType` (Kafka/Cluster/S3/Https/Kinesis/Lambda/SchemaRegistry/Sample), `connectionConfig` |
 | `processor` | `processorName`, `pipeline` (must start with `$source`, end with `$merge`/`$emit`), `dlq`, `autoStart` |
-| `privatelink` | `privateLinkProvider`, `privateLinkConfig` |
+| `privatelink` | `privateLinkConfig` (project-level, not tied to a specific workspace) |
 
 **Field mapping — only fill fields for the selected resource type:**
 
 - **resource = "workspace":** Fill: `projectId`, `workspaceName`, `cloudProvider`, `region`, `tier`, `includeSampleData`. Leave empty: all connection and processor fields.
 - **resource = "connection":** Fill: `projectId`, `workspaceName`, `connectionName`, `connectionType`, `connectionConfig`. Leave empty: all workspace and processor fields. (See [references/connection-configs.md](references/connection-configs.md) for type-specific schemas.)
 - **resource = "processor":** Fill: `projectId`, `workspaceName`, `processorName`, `pipeline`, `dlq` (recommended), `autoStart` (optional). Leave empty: all workspace and connection fields. (See [references/pipeline-patterns.md](references/pipeline-patterns.md) for pipeline examples.)
-- **resource = "privatelink":** Fill: `projectId`, `workspaceName`, `privateLinkProvider`, `privateLinkConfig`. Leave empty: all connection and processor fields.
+- **resource = "privatelink":** Fill: `projectId`, `privateLinkConfig`. Note: PrivateLink is **project-level**, not workspace-level. `workspaceName` is not required — omit it. Leave empty: all connection and processor fields.
 
 ### atlas-streams-manage — ALL update/state operations
 | Action | Notes |
@@ -99,7 +97,7 @@ If the MongoDB MCP Server is not connected or the streams tools are missing, see
 
 - `resource: "workspace"` → `workspaceName`
 - `resource: "connection"` or `"processor"` → `workspaceName`, `resourceName`
-- `resource: "privatelink"` or `"peering"` → `resourceName` (the ID)
+- `resource: "privatelink"` or `"peering"` → `resourceName` (the ID). These are project-level resources, not tied to a specific workspace.
 
 **Before deleting a workspace**, inspect it first:
 1. `atlas-streams-discover` → `inspect-workspace` — get connection/processor counts
@@ -223,11 +221,17 @@ See [references/development-workflow.md](references/development-workflow.md) for
 
 **Debug a failing processor:**
 1. `atlas-streams-discover` → `diagnose-processor` — one-shot health report. Always call this first.
-2. `atlas-streams-discover` → `get-logs` (`logType: "operational"`) — runtime errors, Kafka failures, schema issues, OOM messages. Always call this second.
-3. **Commit to a specific root cause.** Match symptoms to diagnostic patterns (Error 419, idle partitions, OOM, DLQ errors, etc.) — see [references/output-diagnostics.md](references/output-diagnostics.md) for the full pattern table.
-4. Classify processor type before interpreting output volume (alert vs transformation vs filter).
-5. Provide concrete, ordered fix steps specific to the diagnosed root cause.
-6. If lifecycle event history needed → `atlas-streams-discover` → `get-logs`, `logType: "audit"`
+2. **Commit to a specific root cause.** Match symptoms to diagnostic patterns:
+   - **Error 419 + "no partitions found"** → Kafka topic doesn't exist or is misspelled
+   - **State: FAILED + multiple restarts** → connection-level error (bypasses DLQ), check connection config
+   - **State: STARTED + zero output + windowed pipeline** → likely idle Kafka partitions blocking window closure; add `partitionIdleTimeout` to Kafka `$source` (e.g., `{"size": 30, "unit": "second"}`)
+   - **State: STARTED + zero output + non-windowed** → check if source has data; inspect Kafka offset lag
+   - **High memoryUsageBytes approaching tier limit** → OOM risk; recommend higher tier
+   - **DLQ count increasing** → per-document errors; use MongoDB `find` on DLQ collection
+   See [references/output-diagnostics.md](references/output-diagnostics.md) for the full pattern table.
+3. Classify processor type before interpreting output volume (alert vs transformation vs filter).
+4. Provide concrete, ordered fix steps specific to the diagnosed root cause. Do NOT present a list of hypothetical scenarios.
+5. If detailed logs are needed, direct the user to the Atlas UI: **Atlas → Stream Processing → Workspace → Processor → Logs tab**.
 
 ### Chained processors (multi-sink pattern)
 **CRITICAL: A single pipeline can only have ONE terminal sink** (`$merge` or `$emit`). When users request multiple output destinations (e.g., "write to Atlas AND emit to Kafka"), you MUST acknowledge the single-sink constraint and propose chained processors using an intermediate destination. See [references/pipeline-patterns.md](references/pipeline-patterns.md) for the full pattern with examples.
--- a/skills/atlas-stream-processing/SKILL.md
+++ b/skills/atlas-stream-processing/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: atlas-stream-processing
-description: "Manages MongoDB Atlas Stream Processing (ASP) workflows. Requires MongoDB MCP Server with Atlas API credentials. Handles workspace provisioning, data source/sink connections, processor lifecycle operations, debugging diagnostics, and tier sizing. Supports Kafka, Atlas clusters, S3, HTTPS, and Lambda integrations for streaming data workloads and event processing. NOT for general MongoDB queries or Atlas cluster management."
+description: "Manages MongoDB Atlas Stream Processing (ASP) workflows. Handles workspace provisioning, data source/sink connections, processor lifecycle operations, debugging diagnostics, and tier sizing. Supports Kafka, Atlas clusters, S3, HTTPS, and Lambda integrations for streaming data workloads and event processing. NOT for general MongoDB queries or Atlas cluster management. Requires MongoDB MCP Server with Atlas API credentials."
 metadata:
   version: 1.0.0
   user-invocable: "true"
--- a/skills/atlas-stream-processing/SKILL.md
+++ b/skills/atlas-stream-processing/SKILL.md
@@ -35,7 +35,7 @@ If the MongoDB MCP Server is not connected or the streams tools are missing, see
 | `list-processors` | See all processors in a workspace |
 | `inspect-processor` | Check processor state, pipeline, config |
 | `diagnose-processor` | Full health report: state, stats, errors |
-| `get-networking` | PrivateLink and VPC peering details |
+| `get-networking` | PrivateLink and VPC peering details. Optional: `cloudProvider` + `region` to get Atlas account details for PrivateLink setup |
 
 **Pagination** (all list actions): `limit` (1-100, default 20), `pageNum` (default 1).
 **Response format**: `responseFormat` — `"concise"` (default for list actions) or `"detailed"` (default for inspect/diagnose).
@@ -67,7 +67,7 @@ If the MongoDB MCP Server is not connected or the streams tools are missing, see
 
 **Field mapping** — always fill `projectId`, `workspaceName`, then by action:
 
-- `"start-processor"` → `resourceName`. Optional: `tier`, `resumeFromCheckpoint`, `startAtOperationTime`
+- `"start-processor"` → `resourceName`. Optional: `tier`, `resumeFromCheckpoint`, `startAtOperationTime` (ISO 8601 timestamp to resume from a specific point)
 - `"stop-processor"` → `resourceName`
 - `"modify-processor"` → `resourceName`. At least one of: `pipeline`, `dlq`, `newName`
 - `"update-workspace"` → `newRegion` or `newTier`
@@ -79,6 +79,8 @@ If the MongoDB MCP Server is not connected or the streams tools are missing, see
 - `start-processor` → errors if processor is already STARTED
 - `stop-processor` → no-ops if already STOPPED or CREATED (not an error)
 - `modify-processor` → errors if processor is STARTED (must stop first)
+
+**Processor states:** `CREATED` → `STARTED` (via start) → `STOPPED` (via stop). Can also enter `FAILED` on runtime errors. Modify requires STOPPED or CREATED state.
 
 **Teardown safety checks:**
 - **Processor deletion** → auto-stops before deleting (no need to stop manually first)
@@ -144,7 +146,7 @@ Key quickstarts:
 
 See [references/pipeline-patterns.md](references/pipeline-patterns.md) for stage field examples with JSON syntax.
 
-**SchemaRegistry connection:** `connectionType` must be `"SchemaRegistry"` (not `"Kafka"`). See [references/connection-configs.md](references/connection-configs.md#schemaregistry) for required fields and auth types.
+**SchemaRegistry connection:** `connectionType` must be `"SchemaRegistry"` (not `"Kafka"`). Schema type values are case-sensitive (use lowercase `avro`, not `AVRO`). See [references/connection-configs.md](references/connection-configs.md#schemaregistry) for required fields and auth types.
 
 ## MCP Tool Behaviors
 
--- a/skills/atlas-stream-processing/SKILL.md
+++ b/skills/atlas-stream-processing/SKILL.md
@@ -1,0 +1,385 @@
+---
+name: atlas-stream-processing
+description: "Build, operate, and debug MongoDB Atlas Stream Processing (ASP) pipelines, streaming data workloads, and event streams. Covers workspaces, connections, processors, and Kafka. NOT for general MongoDB queries or Atlas cluster management."
+metadata:
+  version: 1.0.0
+  user-invocable: true
+---
+
+# MongoDB Atlas Streams
+
+Build, operate, and debug Atlas Stream Processing (ASP) pipelines using four MCP tools from the MongoDB MCP Server.
+
+## Prerequisites
+
+This skill requires the **MongoDB MCP Server** connected with:
+- Atlas API credentials (`apiClientId` and `apiClientSecret`)
+- `previewFeatures: ["streams"]` enabled in the MCP server config
+
+The 4 tools: `atlas-streams-discover`, `atlas-streams-build`, `atlas-streams-manage`, `atlas-streams-teardown`.
+
+## If MCP tools are unavailable
+
+If the MongoDB MCP Server is not connected or the streams tools are missing:
+1. Verify MCP server config has `previewFeatures: ["streams"]` enabled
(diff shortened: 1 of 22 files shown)


skills/atlas-stream-processing/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+
+This skill requires the **MongoDB MCP Server** connected with:
+- Atlas API credentials (`apiClientId` and `apiClientSecret`)
+- `previewFeatures: ["streams"]` enabled in the MCP server config
+
+The 4 tools: `atlas-streams-discover`, `atlas-streams-build`, `atlas-streams-manage`, `atlas-streams-teardown`.
+
+## If MCP tools are unavailable
+
+If the MongoDB MCP Server is not connected or the streams tools are missing:
```
dacharyc: This section is a good general approach, but we can probably be clearer in surfacing choices to the user here versus just hinting to the agent to use other tools. And this also uses tokens that may not be needed if everything is configured properly. So we could improve token efficiency and user experience if we move this info out to a separate reference file, and refer to it here:

> If the MongoDB MCP Server is not connected or the streams tools are missing, see [references/mcp-troubleshooting.md](references/mcp-troubleshooting.md) for diagnostic steps and fallback options.

The separate file could look something like this - generated with Claude with some structural/conditional logic help from me, but the examples would want to be verified manually: https://gist.github.com/dacharyc/beb1b945fa4ab1aaa732525f3a8422de

skills/atlas-stream-processing/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+
+# MongoDB Atlas Streams
+
+Build, operate, and debug Atlas Stream Processing (ASP) pipelines using four MCP tools from the MongoDB MCP Server.
+
+## Prerequisites
+
+This skill requires the **MongoDB MCP Server** connected with:
+- Atlas API credentials (`apiClientId` and `apiClientSecret`)
+- `previewFeatures: ["streams"]` enabled in the MCP server config
```
dacharyc: Just noting here - I'm guessing `previewFeatures` is a feature flag, so we'll want to update this skill when the feature flag is no longer needed. I'll chat with Betsy and Eliza - we'll need some way to keep track of things like this that we _know_ will need updates in the future. But I don't know if there's anything from a PM side that you normally do to keep track of places that need updates when things change - if so, we'd want to note this as a place that needs updates.

skills/atlas-stream-processing/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+
+If the MongoDB MCP Server is not connected or the streams tools are missing:
+1. Verify MCP server config has `previewFeatures: ["streams"]` enabled
+2. For read-only exploration: use `atlas streams` CLI commands (requires Atlas CLI installed)
+3. For pipeline prototyping: use `sp.process()` in mongosh (runs pipelines ephemerally, no billing)
+4. Full CRUD operations require the MCP server — help the user fix their setup
+
+## Tool Selection Matrix
+
+**Every tool call requires `projectId`.** If unknown, call `atlas-list-projects` first.
```
dacharyc: If all the tool calls require `projectId`, we probably want to move this into the prerequisites along with the Atlas API credential requirement. Also, presumably the "call `atlas-list-projects`" suggestion requires the Atlas CLI to be installed, so we should note that as a prerequisite for the agent to try calling `atlas-list-projects` - possibly with instructions for the agent to surface the missing ID and/or tool info for the user to decide how to proceed.

skills/atlas-stream-processing/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+
+### atlas-streams-discover — ALL read operations
```
dacharyc, line 28: Not a requirement, just flagging for your consideration - do we expect users to be doing all of these operation types every time they invoke the skill? i.e. `read/create/update/delete`? Or do we think it'll be mainly `read`, with `create/update/delete` being secondary activities?

If we think the latter, we could potentially trim these other sections to reference files so they're not loaded into context _every_ time the skill loads, and flag their location for the agent. But if we think the user is likely to need multiple of these operation types every time they use the skill, then the potential token gain at skill load time probably isn't worth maintaining the separate files. It's a relatively small number of tokens in either case, so no hard requirement or strong suggestion here, but these are the types of questions we should be considering when authoring skills.

skills/atlas-stream-processing/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+- `resource: "privatelink"` or `"peering"` → `resourceName` (the ID)
+
+**Before deleting a workspace**, inspect it first:
+1. `atlas-streams-discover` → `inspect-workspace` — get connection/processor counts
+2. Present to user: "Workspace X contains N connections and M processors. Deleting permanently removes all. Proceed?"
+3. Wait for confirmation before calling `atlas-streams-teardown`
+
+## CRITICAL: Validate Before Creating Processors
+
+**You MUST call `search-knowledge` before composing any processor pipeline.** This is not optional.
```
dacharyc, line 111: If this is only needed for the `create` operations, and we would potentially consider moving the `create` operation out to a separate file, we could place this in the `create` reference file and omit it from the token s loaded here. Since it's only needed conditionally, it may not make sense to include it in the base SKILL.md file.

skills/atlas-stream-processing/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+Key quickstarts:
+| Quickstart | Pattern |
+|-----------|---------|
+| `00_hello_world.json` | Inline `$source.documents` with `$match` (zero infra, ephemeral) |
+| `01_changestream_basic.json` | Change stream → tumbling window → `$merge` to Atlas |
+| `03_kafka_to_mongo.json` | Kafka source → tumbling window rollup → `$merge` to Atlas |
+| `04_mongo_to_mongo.json` | Chained processors: rollup → archive to separate collection |
+| `05_kafka_tail.json` | Real-time Kafka topic monitoring (sinkless, like `tail -f`) |
+
+## Pipeline Rules & Warnings
```
dacharyc, line 126: If these are all only needed by `create`, too, then it probably does make sense to move this out to a dedicated `create` reference file. Seems like several sections that may only pertain to the one operation type.

skills/atlas-stream-processing/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+- **`$externalFunction`**: must include `connectionName`, `functionName`, `execution`, `as`, `onError: "dlq"`
+- **`$validate`**: must include `validator` with `$jsonSchema` and `validationAction: "dlq"`
+- **`$lookup`**: include `parallelism` setting (e.g., `parallelism: 2`) for concurrent I/O
+- **AWS connections** (S3, Kinesis, Lambda): IAM role ARN must be registered via Atlas Cloud Provider Access first. Always confirm this with user. See [references/connection-configs.md](references/connection-configs.md) for details.
+
+See [references/pipeline-patterns.md](references/pipeline-patterns.md) for stage field examples with JSON syntax.
+
+**SchemaRegistry connection:** `connectionType` must be `"SchemaRegistry"` (not `"Kafka"`). See [references/connection-configs.md](references/connection-configs.md#schemaregistry) for required fields and auth types.
+
+## MCP Tool Behaviors
```
dacharyc, line 151: It seems this content is mainly needed with `create` operations, although some of it also applies to `update`. If we're moving `create` stuff out to a `create` reference file, I'd move these things there, and consider duplicating the ones that are relevant to `update` in a `update` reference file.

skills/atlas-stream-processing/references/sizing-and-parallelism.md
```diff
@@ -0,0 +1,180 @@
+# Sizing & Parallelism Reference
+
+Adapted from [kgorman/asp_claude](https://github.com/kgorman/asp_claude) sizing and parallelism guides, translated for the MongoDB MCP Server streams tools.
```
dacharyc: Just noting here that since this is adapted from that other content, we may need to track changes to that other content to consider when changes here are needed. 

I'll also flag this for Eliza and Betsy, but something to keep in mind for your team when considering things that need updating as guidance evolves.

skills/atlas-stream-processing/references/pipeline-patterns.md
```diff
@@ -0,0 +1,431 @@
+# Pipeline Patterns Reference
+
+**Official examples repo**: https://github.com/mongodb/ASP_example
+Always consult the official repo for the latest validated patterns before creating processors.
```
dacharyc, line 4: Can we be clearer about where in the repo to refer for "latest validated patterns"? The more specificity we can provide, the better the chances that the agent can find and apply relevant patterns from that repo. If we can't be specific here, and instead send an agent out to that repo to "look around" and try to figure out what applies, it may fail to find relevant info, it may find info and mis-apply it, or it may find the correct info and apply it correctly. In any event, not specifying means token use/cost to the developer - if there are reasons for us to make that tradeoff, that's fine, but we should be clear that it's a tradeoff.

skills/atlas-stream-processing/references/output-diagnostics.md
```diff
@@ -0,0 +1,153 @@
+# Processor Output Diagnostics Reference
+
+Adapted from [kgorman/asp_claude](https://github.com/kgorman/asp_claude) output patterns guide, translated for the MongoDB MCP Server streams tools.
```
dacharyc: Flagging this similar to the others - external source that may receive updates independent of this one, and therefore we may need to carry updates back here when that source is updated.

skills/atlas-stream-processing/references/development-workflow.md
```diff
@@ -0,0 +1,305 @@
+# Development Workflow Reference
+
+Adapted from [kgorman/asp_claude](https://github.com/kgorman/asp_claude) development workflow, translated for the MongoDB MCP Server streams tools.
```
dacharyc: Flagging this similar to the others - external source that may receive updates independent of this one, and therefore we may need to carry updates back here when that source is updated.

skills/atlas-stream-processing/references/connection-configs.md
```diff
@@ -0,0 +1,269 @@
+# Connection Configuration Reference
+
+**Official examples repo**: https://github.com/mongodb/ASP_example
```
dacharyc: Do we want an agent to take some action from this information? Are we expecting it to go out and look at this content? Are we providing it as a reference if the agent needs additional examples? We should probably add some information here about when we want the agent to follow this link, and what it should do after it does.

dacharyc (commented):
Hey Laura - thanks for putting this together! It evaluates well with the `review-skill` skill and it's obvious you've put effort into making this a useful resource. 🎉 

I've got a handful of comments here - things for us to flag as "may need updates when external sources get updated," and some possible things we could move out of SKILL.md to improve token efficiency. There's also a suggestion to move some MCP troubleshooting info out to a separate file to provide more details for troubleshooting, and avoid putting those tokens into the happy-path run of this skill.

I did not look at the evals, so I can't speak to that, but overall the content looks sound and I don't think any of my suggestions are blocking - just some improvements we could make.

skills/atlas-stream-processing/SKILL.md
laurazhukas: Updated the prerequisites. 

The atlas-lists-projects MCP tool seems to be based on the api rather than the cli so I dont believe that is required in this case.

skills/atlas-stream-processing/references/sizing-and-parallelism.md
laurazhukas: Oh, the content is not pulling from that link. We used the resources there to build out the initial skill before iterating / flushing it out here. I will remove the link to avoid confusion

skills/atlas-stream-processing/references/output-diagnostics.md
laurazhukas: Same response as the one on your initial comment - We used the resources there to build out the initial skill before iterating / flushing it out here. I will remove the link to avoid confusion

skills/atlas-stream-processing/references/development-workflow.md
laurazhukas: Same response as the one on your initial comment - We used the resources there to build out the initial skill before iterating / flushing it out here. I will remove the link to avoid confusion

skills/atlas-stream-processing/SKILL.md
laurazhukas: Good point, I've created a ticket for removing this in the future (STREAMS-2323)

skills/atlas-stream-processing/SKILL.md
jwongmongodb: Just a correction, the `atlas-list-projects` is a MCP tool [here](https://github.com/mongodb-js/mongodb-mcp-server/blob/67c929c19c97cc5b4540188d10d3374df943eb05/src/tools/atlas/read/listProjects.ts#L9). The idea is for the skill to call the list projects MCP tool if the user does not supply the projectID.

skills/atlas-stream-processing/references/connection-configs.md
laurazhukas: updated

skills/atlas-stream-processing/references/pipeline-patterns.md
laurazhukas: updated it to check the specific folder relevant to this md

skills/atlas-stream-processing/SKILL.md
laurazhukas: added

skills/atlas-stream-processing/SKILL.md
laurazhukas: You bring up a good consideration. Part of the use cases is discovery / development so I anticipate a closer split between read and create. The suggested approach would be a large reduction in the skills file, so I think it is an optimization worth considering in the future. Previously when testing with the skill-reviewer additions / deletions into the skill.md had notable effect on the scores, so it is a change I'd prefer to save for the future.

skills/atlas-stream-processing/SKILL.md
laurazhukas: related to comment above https://github.com/mongodb/agent-skills/pull/10#discussion_r2921005013

dacharyc (approved):
Thanks for making the updates where it made sense, @laurazhukas ! LGTM on the quality checks once the new stream processing tools are available in the MCP server. ✅

skills/atlas-stream-processing/SKILL.md
```diff
@@ -0,0 +1,276 @@
...
+| Action | Use when |
+|--------|----------|
+| `list-workspaces` | See all workspaces in a project |
+| `inspect-workspace` | Review workspace config, state, region |
+| `list-connections` | See all connections in a workspace |
+| `inspect-connection` | Check connection state, config, health |
+| `list-processors` | See all processors in a workspace |
+| `inspect-processor` | Check processor state, pipeline, config |
+| `diagnose-processor` | Full health report: state, stats, errors |
+| `get-logs` | Operational logs (runtime errors) or audit logs (lifecycle) |
```
doug-rodgers-mongodb: Will get-logs expose any secrets, credentials, tokens, or PII? Can we prevent that from happening or tell the LLM to use an option that can reduce the exposure?

skills/atlas-stream-processing/references/connection-configs.md
```diff
@@ -0,0 +1,269 @@
...
+3. Verify connection name matches actual target (clusterName, bootstrapServers, url, etc.)
+4. If mismatch exists, consider renaming or warn the user
+```
+
+See [development-workflow.md](development-workflow.md) "Pre-Deployment Connection Validation" section for the complete validation procedure.
+
+## Important Notes
+- HTTPS connections are for `$https` enrichment ONLY — they are NOT valid as `$source` data sources
+- Store API authentication in connection settings, never hardcode in processor pipelines
+- AWS connections (S3, Kinesis, Lambda) require IAM role ARN registered via Atlas Cloud Provider Access first
```
doug-rodgers-mongodb, line 72: Should we recommend using a dedicated, least-privilege role scoped only to the buckets/streams/functions required by this processor

skills/atlas-stream-processing/SKILL.md
jwongmongodb: Yes, getLog can expose sensitive data. The raw log entries can contain:  
- User-controlled values in error messages and attributes, which may indirectly contain PII or application data.
- Connection and resource metadata such as tenant/project/workspace/processor identifiers, hostnames, etc.
- Query or pipeline fragments in logged validation errors or diagnostics, which can reference sensitive fields or values.

I don't believe that there is redaction for the current logs endpoint. 
We can remove getLogs from our skills and mcp-tools in the meantime until we have better redaction.

skills/atlas-stream-processing/references/connection-configs.md
jwongmongodb: Yes, that makes sense and follows AWS recommendation for least privilege access, good catch

Here is what I will use
```
AWS connections (S3, Kinesis, Lambda) require IAM role ARN registered via Atlas Cloud Provider Access first. Use a dedicated, least-privilege role scoped to only the specific buckets, streams, or functions required by the processor.
 ```

**Comment by jwongmongodb** [comment]
@dacharyc, can you take another look at the latest [commit](https://github.com/mongodb/agent-skills/pull/10/changes/6e0e46d12d27bc0ed28f25fac69c958f9b7ff8b6) 

I've added some improvements:

- Added more details on private link guidance in the references
- I've removed the references to getLogs due to @doug-rodgers-mongodb's concern about the PII
- I also corrected the region information to be the correct Streams regions used in our API (These came up in my MCP tool tests). 

I ran the review-skills with bedrock and got the following:
Screenshot 2026-03-20 at 3 59 02 PM


Novelty Assessment — EXCELLENT (mean: 4.86)

  Well above the threshold of 3. The LLM judge identified genuinely novel content in every file:
  - Proprietary MCP tool signatures (atlas-streams-discover, atlas-streams-build, etc.)
  - Streams-specific region naming (VIRGINIA_USA, eastus, US_CENTRAL1)
  - ASP-specific constraints ($$NOW unavailable, single-sink-per-pipeline, partitionIdleTimeout)
  - Internal workflow enforcement (pre-deployment validation protocol, tier sizing formula)
  - Undocumented gotchas (error 419, $validate with validationAction: "error" crashes)

File: skills/atlas-stream-processing/references/connection-configs.md (hunk 14 above)
Comment (jwongmongodb) [reply skills/atlas-stream-processing/references/connection-configs.md:72]: this is added now

File: skills/atlas-stream-processing/SKILL.md (hunk 13 above)
Comment (jwongmongodb) [reply skills/atlas-stream-processing/SKILL.md]: I've removed the reference to getLogs until we can redact things correctly from the API/MCP tools

**Review by dacharyc — APPROVED** [review]
Hey @jwongmongodb - thanks for your updates! I've taken a look and it seems like these are good factual additions that don't introduce any new considerations I would flag. I think we're good to merge whenever the MCP server release makes the new tools available. ✅

**Comment by jwongmongodb** [comment]
@dacharyc can you take a look at the latest 2 commits, the first one removes the streams feature flag usage with the MCP server and any other stale references to the get-logs endpoint (that was removed previously). The second commit is a minor improvement that came from manual testing. The mcp server has released so I can merge this skill right after approval.

The review-skill got the same score  4.83 when I reran it.

File: skills/atlas-stream-processing/SKILL.md [hunk 15]
```diff
@@ -223,11 +221,17 @@ See [references/development-workflow.md](references/development-workflow.md) for
 
 **Debug a failing processor:**
 1. `atlas-streams-discover` → `diagnose-processor` — one-shot health report. Always call this first.
-2. `atlas-streams-discover` → `get-logs` (`logType: "operational"`) — runtime errors, Kafka failures, schema issues, OOM messages. Always call this second.
-3. **Commit to a specific root cause.** Match symptoms to diagnostic patterns (Error 419, idle partitions, OOM, DLQ errors, etc.) — see [references/output-diagnostics.md](references/output-diagnostics.md) for the full pattern table.
-4. Classify processor type before interpreting output volume (alert vs transformation vs filter).
-5. Provide concrete, ordered fix steps specific to the diagnosed root cause.
-6. If lifecycle event history needed → `atlas-streams-discover` → `get-logs`, `logType: "audit"`
+2. **Commit to a specific root cause.** Match symptoms to diagnostic patterns:
+   - **Error 419 + "no partitions found"** → Kafka topic doesn't exist or is misspelled
```
dacharyc, line 225: One question here, @jwongmongodb - do you have any evals around including this info here? I have observed in other testing that sometimes when you get specific with details like this, and _also_ ask an agent to refer to some supplemental material, sometimes the agent treats a specific list like this with more than one or two entries as a comprehensive reference and does not bother to access the secondary material. I'm wondering if we have any data driving these details being included here versus telling the agent to refer to the `output-diagnostics` reference for _all_ failure details.

In other words, I'm wondering if this should be:

```
2. **Commit to a specific root cause.** Match symptoms to diagnostic patterns detailed in [references/output-diagnostics.md](references/output-diagnostics.md)
```

And just omitting details from here at all. Or if maybe you have found this is needed because evals indicate the agent is less likely to retrieve or correctly apply the details when they're in a secondary reference.

dacharyc (approved):
Back to you with one minor question. 

Overall this looks good and this question isn't blocking - it's just a small Q about whether we may want to handle diagnostic information differently.

skills/atlas-stream-processing/SKILL.md
jwongmongodb: I ran some a/b testing evals for the following 3 prompts:
Screenshot 2026-03-24 at 3 07 36 PM

and the inlined version was noticeable better:
Screenshot 2026-03-24 at 3 07 05 PM

```
The reviewer's concern is valid in general but doesn't apply here. Here's why, based on the actual run data:

  The reviewer's model: "Inline list → agent treats it as exhaustive → skips reference"

  This is true for enumeration lists — if you inline "supported formats: JSON, CSV, Parquet" the agent may not check for more. But the debugging patterns aren't an enumeration.
   They're a conditional decision tree: "IF symptom X THEN root cause Y." The agent doesn't need to enumerate all patterns — it needs to match the one that fits.

  What actually happened with Version B (reference pointer only)

  The Version B agents didn't read output-diagnostics.md at all. In 8 of 9 runs, the agent went straight from diagnose-processor to guessing based on general streaming
  knowledge. The reference pointer ("detailed in references/output-diagnostics.md") wasn't sufficient motivation to fetch and read a separate file during a debugging task.

  This is because debugging is time-pressured reasoning — the agent wants to give an answer now, not go read a 139-line reference file. When the patterns are inline, they're
  already in context and get matched immediately.

  Why the inline patterns don't cause the "exhaustive list" problem

  1. They're conditional, not enumerative. The agent matches ONE pattern based on symptoms — it doesn't need to consider all 6. There's no "am I missing one?" instinct like
  there would be with a list of options.
  2. The reference file serves a different purpose. The inline patterns handle root cause identification. The reference file handles the deeper workflow: processor type
  classification, DLQ inspection procedures, contextual factors (time of day, seasonality). These are separate concerns, not overlapping lists.
  3. The evidence shows no "crowding out." Version A agents that had the inline patterns AND the reference pointer didn't perform worse on any dimension. They got 100% on the
  specific diagnosis AND still referenced deeper concepts (processor type classification) when relevant.

  The real tradeoff

  The reviewer framed it as "inline details vs reference file." But the data shows it's actually "something vs nothing" — Version B agents didn't use either resource
  effectively. The choice isn't between two access patterns; it's between having the critical patterns in context or hoping the agent will go fetch them (it won't, 89% of the
  time).
  ```

jwongmongodb: can address this in the next PR to enhance/tweak these things.

PR #11, opened 2026-03-09
Rework skill validator script

nirinchev:
This PR proposes the following changes:
* Moves `validate-skills.sh` script to `tools` so that it can more easily be accessed outside of github actions.
* Tweaks `validate-skills.sh` behavior:
  * Removes changed file logic - the script already runs very quickly and passing structural validation is a prerequisite for merging PRs to main, making the logic to only run on changed files unnecessarily complex for the value we're getting.
  * Adds validation that `skill-validator` is installed.
  * Decouple reporting from GHA, so that it can be used locally.
* Stub a basic CONTRIBUTING.md with instructions how to run the validate-skills script as well as the `review-skill` skill.


nirinchev:
I'm going to go ahead and merge this as there's a few other PRs that are blocked on it, but feel free to do a comprehensive review and I'll address the comments in a follow-up.

PR #14, opened 2026-03-10
Add release workflows MCP-430

nirinchev:
Adds release workflows as well as codeql workflow that analyzes our GHA workflows.

Release flow is:
1. Trigger draft-release workflow with patch | minor | major argument or an exact version to release.
2. Workflow updates versions and creates a PR.
3. On each push to main, run the draft release workflow that checks if the current version in plugin.json and friends has been released.
4. If not released, creates a draft release and generates release notes.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,26 @@
+name: CodeQL (GitHub Actions)
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  actions: read
+  contents: read
+  security-events: write
+
+jobs:
+  analyze-actions:
+    name: Analyze workflows
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v4
+        with:
+          languages: actions
+
+      - name: Perform CodeQL analysis
+        uses: github/codeql-action/analyze@v4
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -1,0 +1,66 @@
+name: Draft Release
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - ".cursor-plugin/plugin.json"
+      - ".claude-plugin/plugin.json"
+      - "gemini-extension.json"
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  draft-release:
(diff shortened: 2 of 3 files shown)


dacharyc (commented):
The workflow mechanics look sound, but I have a few questions about the release model before approving:

1. What do versions track? The release workflow only bumps versions in the plugin JSON files, and only triggers when those files change — but most of the work in this repo is skill content (markdown), which can (will?) ship without a version bump. What's the version meant to represent?
2. Will the generated release notes be useful? `--generate-notes` produces a flat list of PR titles since the last tag. Since there's no .github/release.yml to categorize PRs, and skill changes won't trigger a release on their own, the notes may not meaningfully reflect what's actually shipped.
3. What does semver mean here? What constitutes major vs. minor vs. patch for a plugin that's mostly markdown skills plus MCP config?

nirinchev:
I'll prefix the answers with calling out that these versions only impact users who have installed the plugin from a marketplace and have no bearing for someone who just cloned the repo and added the skills to their agent. With that in mind, here are my thoughts:

1. The plugin.json version is used by marketplaces to notify users that there's an update to a plugin they have installed. On most platforms, the published plugin does not get automatically updated from the repo and instead we need to submit it for approval ([Cursor docs](https://cursor.com/help/security-and-privacy/marketplace-security#are-plugin-updates-reviewed)). On others, such as gemini, the marketplace indexes github releases and picks up the new version automatically ([docs](https://geminicli.com/docs/extensions/releasing/#release-through-github-releases)). In all cases though, if we update the skills without bumping the version, users wouldn't be prompted to update their installation.

2. The generated release notes are a starting point and the person doing the release can (and should) edit them before publishing. We can look for more comprehensive solutions in the future, but this feels like a good enough first step. It still involves some manual work, but it's better than nothing.

3. Semver is how we communicate to the plugin user what they can expect from the new version and we should probably try and stick as much as possible to the spirit of the spec without necessarily being overly pedantic about it. From that perspective, patch versions would be minor tweaks to existing workflows - e.g. some added validation or removing ambiguity - essentially, no changes to the functionality provided by the plugin, but improvements in the guidance provided by the agent. Minor versions could be new skills being added or substantial enhancements to existing ones (e.g. new dimension for offering data modeling advice). And major releases would involve removing functionality or changes that would otherwise break legitimate user workflows.

dacharyc (approved):
Sounds reasonable. I appreciate capturing the thinking here for future reference. I think the industry has big-picture questions to figure out around versioning skill content, making updates visible, etc. but that is largely out of scope of these things.

PR #23, opened 2026-03-16
Modify the setup skill to avoid credential handling

nirinchev:
This modifies the instructions in the `mongodb-mcp-setup` skill to eliminate the need for credential passing into the LLM context.

### DIFF
--- a/skills/mongodb-mcp-setup/SKILL.md
+++ b/skills/mongodb-mcp-setup/SKILL.md
@@ -35,7 +35,7 @@ Before starting the setup, check if the user already has the required environmen
 Run this command to check for existing configuration (masking values to avoid exposing credentials):
 
 ```bash
-env | grep -E "^MDB_MCP_(CONNECTION_STRING|API_CLIENT_ID|API_CLIENT_SECRET|READ_ONLY)=" | sed 's/=.*/=[set]/'
+env | grep "^MDB_MCP" | sed 's/=.*/=[set]/'
 ```
 
 **Interpretation:**
@@ -282,5 +282,5 @@ Proceed to Step 6 (Next Steps).
 - **Client doesn't pick up variables**: Ensure full restart (quit + reopen), not just a reload
 - **Invalid connection string format**: Re-check the format; must start with `mongodb://` or `mongodb+srv://`
 - **Atlas Admin API errors (Option B)**: Verify your IP is in the service account's API Access List
-- **Read-only mode not working**: Check `MDB_MCP_READ_ONLY=true` is set — run `env | grep MDB_MCP`
+- **Read-only mode not working**: Check `MDB_MCP_READ_ONLY=true` is set — run `env | grep ^MDB_MCP_READ_ONLY`
 - **fish/PowerShell**: Syntax differs — use `set -x` (fish) or `$env:` (PowerShell) instead of `export`
--- a/skills/mongodb-mcp-setup/SKILL.md
+++ b/skills/mongodb-mcp-setup/SKILL.md
@@ -26,83 +26,80 @@ The MongoDB MCP server requires authentication. Users have three options:
    - Runs Atlas locally in Docker, requires Docker installed
    - No credentials or cloud cluster access
 
-**Important**: Code snippets use bash/zsh syntax. For other shells (PowerShell, fish, etc.), adjust commands to the user's shell.
-
-## Execution Modes
-
-This skill supports two execution modes:
-
-1. **Interactive Mode** (preferred): Directly modifies the user's shell profile when Bash access is available
-2. **Documentation Mode** (fallback): Creates example configuration files and instructions when Bash access is restricted
-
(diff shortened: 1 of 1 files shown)


dacharyc (approved):
Solid wins here; the credential handling removal simplifies the flow, and consolidating the troubleshooting/error sections addresses the token efficiency concerns from the last round. Looks good.

One minor gap: Steps 1 and 5.1 assume the agent can run shell commands, but there's no fallback if it can't. The old Documentation Mode covered this (albeit imperfectly). A single line like "If the agent cannot run shell commands, ask the user for their shell type and profile path, then skip directly to Step 5.2" would close the loop. Not a blocker; just worth noting for a future pass.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -26,125 +26,115 @@ The MongoDB MCP server requires authentication. Users have three options:
...
 
 ## Step 1: Check Existing Configuration
 
 Before starting the setup, check if the user already has the required environment variables configured.
 
-**Try to run** this command to check for existing configuration:
+Run this command to check for existing configuration:
 
 ```bash
 env | grep -E "MDB_MCP_(CONNECTION_STRING|API_CLIENT_ID|API_CLIENT_SECRET|READ_ONLY)"
```
doug-rodgers-mongodb: Won't this expose the credential to the LLM after it was just told that it wouldn't handle any credentials on line 29?

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -26,125 +26,115 @@ The MongoDB MCP server requires authentication. Users have three options:
...
+Steps:
 
-To configure:
-
-1. On the service account details page, find "API Access List"
-2. Click "Add Access List Entry"
-3. Add current IP or 0.0.0.0/0 for testing (not for production)
+1. On the service account details page, find **API Access List**
+2. Click **Add Access List Entry**
+3. Add current IP address (or `0.0.0.0/0` for testing only — not for production)
```
doug-rodgers-mongodb: I know that you have stated that 0.0.0.0/0 is for testing only, but I would emphasize that using it is a security concern. The end user needs to know that it is a risk and probably not necessary.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -183,171 +173,90 @@ Ask whether they want read-only or read-write access:
...
-**If Bash is denied:** Switch immediately to Documentation Mode (Step 5b).
-
-Based on the shell, determine the appropriate profile file to update from training data. If necessary, refer to the shell docs.
+Based on the result, identify the appropriate profile file using your training data. If unsure which shell or profile they are using, ask them to specify the path.
 
-#### 5a.2: Add Environment Variables
+### 5.2: Show the Exact Snippet to Add
 
-Check if variables already exist in the profile file. If so, replace rather than duplicate.
+Tell the user to open their profile file (e.g. `code ~/.zshrc`, `nano ~/.zshrc`, or their preferred editor) and add the following lines at the end. Make sure to adapt the profile path in the instructions to match their shell.
```
doug-rodgers-mongodb: The shell rc files are normally group and world readable by default most of the time. I think it might be better guidance to would be to add the connection strings into a dedicated ~/.mcp-env file (chmod 600) and then source that file from the shell rc file. This would help protect accidental perm changes to the shell rc file. This would also help prevent users accidentally committing the connection string to git if it was in their shell rc file.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -183,171 +173,90 @@ Ask whether they want read-only or read-write access:
...
-## What to do
-
-Add this to your shell profile (e.g., `~/.zshrc`/`~/.bashrc`/`$PROFILE`):
-
-For Connection String (Option A):
+**Verify the variables are set:**
 
 ```bash
-export MDB_MCP_CONNECTION_STRING="<paste-your-connection-string-here>"
+env | grep MDB_MCP
```
doug-rodgers-mongodb: We are exposing the credentials to the LLM again after stating that they wouldn't be.

skills/mongodb-mcp-setup/SKILL.md
```diff
@@ -183,171 +173,90 @@ Ask whether they want read-only or read-write access:
...
-export MDB_MCP_API_CLIENT_SECRET="<paste-your-client-secret-here>"
-```
+Expected output should show the variable(s) they just added. If nothing appears, check that the profile file path is correct and the file was saved.
 
-**If you chose read-only access**, also add:
+**Security reminder**: Ensure the profile file permissions are restrictive:
 
 ```bash
-export MDB_MCP_READ_ONLY="true"
+chmod 600 ~/.zshrc   # adjust path as needed
```
doug-rodgers-mongodb: See note above about using a dedicated ~/.mcp-env file.

PR #26, opened 2026-03-23
docs: update README TODOs

cbullinger:


gagik (approved):

dacharyc (approved):
Thanks for filling this out, @cbullinger ! Let's :shipit: to get rid of the placeholders since Gemini is now driving traffic here, and we can always refine as needed.

PR #30, opened 2026-03-30
DOCSP-58329: update docs links

cbullinger:
Update / add docs links. 
> NOTE: Cursor link was already correct; Gemini doesn't have any URL parameter to update.


dacharyc (approved):
Thanks for adding the links and fixing up the Claude Plugin installation instructions, @cbullinger ! This LGTM. Let's hang onto it for a couple more hours and then merge as discussed.

PR #33, opened 2026-04-17
chore(schema-design): add schema versioning eval MCP-467

Anemy:
MCP-467

We're in the process of adding evals for the mongodb-schema-design skill. This is the first that is testing an existing schema, and uses the pattern-schema-versioning reference. More evals will be added in subsequent prs. 

I've run this a few separate times, on a run of 10 it passed 10/10 with the skill and 3/10 without (counted 2 partials as 0.5 each). That's using an LLM as a judge. I've only read through a few of the results but it seems to make sense. Once we have more of the evals I'll do a more comprehensive analysis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 # User-specific Claude settings
 .claude/settings.local.json
 .score_cache/
+
+.DS_Store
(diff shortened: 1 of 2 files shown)


paula-stacho (approved):
Looks good! I thought we'd use multiple files instead of a single evals file, but we can change that later.

PR #34, opened 2026-04-22
chore(schema-design): add evals for skill boundary and skill activation MCP-474

Anemy:
MCP-474 

Adds an activation eval, with 40 tests, and a script to run it.
Adds a skill boundary for query-optimizer vs schema-design.

I ran the script with the `trigger-eval.json` in the `mongodb-natural-language-querying` skill and got a 20/20 result.

Example run from the script:
```
Results: 38/40 passed, 2 failed
Total cost:  $0.7797

  True positives:  25   (correctly triggered)
  True negatives:  13   (correctly skipped)
  False positives: 0   (triggered when shouldn't)
  False negatives: 2   (missed when should trigger)

  Precision: 100.0%
  Recall:    92.6%
  F1 Score:  96.2%

Failures:
  [19] MISSED (false negative)
       "How can I optimize this aggregation pipeline that has three $lookup stages?"
       Skills called: [mongodb-query-optimizer]
  [20] MISSED (false negative)
       "Why are my write operations so slow on this collection?"
       Skills called: [mongodb-query-optimizer]

Results written to trigger-eval-results.json
```
The 2 false negatives from the run are a bit ambiguous, I think the query-optimizer could do a fine job on those potentially, although it is part of the skill boundary.

I've run the skill boundaries eval 3 times, it gets 100% on everything except the description gaps. The description gaps vary in each, all above 50% though.

### DIFF
--- a/testing/skills-boundaries/README.md
+++ b/testing/skills-boundaries/README.md
@@ -35,15 +35,6 @@ Tests the boundary between:
 # 4. Mark as pass/fail
 ```
 
-### Automated Testing (Future)
-
-```bash
-# When you have automated skill invocation testing:
-npm run test:skill-boundaries
-# or
-python test_skill_boundaries.py
-```
-
 ## Test Case Structure
 
 ```json
--- a/testing/skills-boundaries/README.md
+++ b/testing/skills-boundaries/README.md
@@ -17,33 +17,17 @@ Tests the boundary between:
 - **mongodb-natural-language-querying**: Standard queries, filtering, aggregation, basic data retrieval
 - **search-and-ai**: Atlas Search, vector search, fuzzy matching, semantic similarity, full-text search
 
-**25 test cases covering:**
-- 5 clear mongodb-natural-language-querying cases (basic filtering, aggregation, SQL translation)
-- 8 clear search-and-ai cases (fuzzy matching, semantic search, full-text search, autocomplete)
-- 8 ambiguous cases (the critical gray zone)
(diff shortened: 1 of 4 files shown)


paula-stacho:
I agree that the last 2 are ambiguous, they fall exactly in the space where we might need either of the two skills. As actually ensuring that both are considered is something that's out of scope now, I've created this ticket to at least collect examples, as this is a first step we've agreed on: https://jira.mongodb.org/browse/MCP-477
I propose to leave the ambiguous prompts out for the time being and adding them as a comment to this ticket, what do you think?

testing/skills-boundaries/README.md
```diff
@@ -23,6 +23,28 @@ Tests the boundary between:
 - 8 ambiguous cases (the critical gray zone)
 - 4 edge cases (optimization, exact match, case-insensitive, SQL translation)
 
+### query-optimizer-vs-schema-design.json
+
+Tests the boundary between:
+- **mongodb-query-optimizer**: Query/index optimization, slow query analysis, explain plans, Performance Advisor
+- **mongodb-schema-design**: Data modeling, embed vs reference, schema patterns/anti-patterns, migrations
+
+**40 test cases covering:**
```
paula-stacho: Are we sure we want these numbers documented? Looks like they could easily get out of sync 😅

Anemy: They were already out of sync, the two updates to this file were by llms in the past I think. I'll remove it them.

paula-stacho (approved):

PR #37, opened 2026-04-23
feat(schema-design): Access pattern analysis MCP-476

paula-stacho:
# MongoDB Agent Skill Update

## Goals

- In addition to the system profiler, the mongodb-schema-design skill includes guidance to utilize Atlas specific workload sources (if available) - in particular the $queryStats and slow query logs (performance advisor)
- Agent can retrieve relevant context from MongoDB Atlas, via MCP or other available tools.
- The agent brings relevant sources to awareness and works with the user to consider these instead of jumping straight into analysis

**NEW**: 
More cautious on mentions of the system profiler:
- All mentions say that other sources are preferred
- Removed system profiler from the main list of sources in SKILL.md, but can still be reached through the `Verify with` sections.
- All examples are moved to the single reference file, so that the cautionary messaging at the beginning of the file isn't missed.

## Author Self-Validation

- [ ] Ran `review-skill` skill to self-review structure, format, content quality, and novelty
- [ ] Tested skill with LLMs using the five prompts above

## SME Review

**SME:** @


skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,54 @@ Reference these guidelines when:
...
+- Codebase - Examine actual queries in application code to understand access patterns, especially for new applications or with changing workloads. Can be used in conjunction with query stats for a more complete picture.
+- Natural language input - Ask the user to describe their typical queries and access patterns in natural language. Can be used as the only source or to supplement and validate other sources - the user might have contextual knowledge that is not reflected in the data or codebase.
+
+**Combining Query Stats and Slow Query Logs:**
+
+Use both together for comprehensive analysis:
+1. Query Stats → identify frequent access patterns (which queries run most often)
+2. Slow Query Logs → identify performance bottlenecks (which queries are slow)
+3. Focus schema optimization on queries that are both frequent AND slow (highest impact)
```
Anemy, line 90: Would it make sense to move the M10+ required items to a specific reference page? It would reduce the context for the general users here then.

skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,54 @@ Reference these guidelines when:
...
+
+#### What to Look For
+
+When analyzing access patterns, identify:
+  - **Co-accessed fields** - Which fields are queried or updated together
+  - **Co-accessed collections** - Which collections are queried or updated together
+  - **Query frequency** - High-frequency queries should drive schema design decisions
+	- **Different update patterns** - If certain fields are updated much more frequently than others, it may make sense to separate them to avoid unnecessary document updates
+  - **Array growth** - Unbounded arrays that grow over time (risk hitting 16MB limit)
```
Anemy: We mention this in the [Schema Fundamentals Summary](https://github.com/mongodb/agent-skills/blob/c459beb3faf843dcf1b876996744f29aa093219a/skills/mongodb-schema-design/SKILL.md#schema-fundamentals-summary) below. I'm wondering if there's a way to reduce the duplication. I'm not sure if this section `What to Look For` is giving the model more actionable information. Is it something we could omit entirely?

skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,54 @@ Reference these guidelines when:
...
+Ask the user:
+  - Is this a new design or is there a production database with existing access patterns to analyze?
+  - If there is production data, is it on Atlas? If yes, what tier? (M0/M2/M5 vs M10+)
+
+**Step 2: Determine workload type**
+Is the workload read-heavy, write-heavy, or balanced? This will influence which diagnostic sources are most relevant.
+Ask the user:
+  - Is your application primarily reading data (analytics, reports, searches) or writing data (logging, IoT ingestion, frequent updates)?
+
+Verify with `db.serverStatus().opcounters`.
```
Anemy, line 70: Should we add the serverStatus with a mention of the opcounters in the sources section?
In the serverStatus would `catalogStats` also be an indicator? It might show they have a ton of collections or databases which may be an anti-pattern. https://www.mongodb.com/docs/manual/reference/command/serverStatus/#catalogstats

skills/mongodb-schema-design/SKILL.md
paula-stacho: I'm not sure. We did say that the focus for now would be atlas users, and I think this becomes even more prominent now that we're removing system.profile. Let's chat more on the sync today.

skills/mongodb-schema-design/SKILL.md
paula-stacho: Fair point, it is reiterating what is said elsewhere. I will try to verify if omitting it entirely produces the same results, alternatively the section could be more concise (list without details)

skills/mongodb-schema-design/SKILL.md
paula-stacho: I'm not sure. On it's own the count is isn't definitive information, it would just be an indicator that we might wanna look into the access patterns - but if we got to this point we're already doing that.

skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,45 @@ Reference these guidelines when:
...
+
+**Step 1: Assess the environment**
+Ask the user:
+  - Is this a new design or is there a production database with existing access patterns to analyze?
+  - If there is production data, is it on Atlas? If yes, what tier? (M0/M2/M5 vs M10+)
+
+**Step 2: Determine workload type**
+Is the workload read-heavy, write-heavy, or balanced? This will influence which diagnostic sources are most relevant.
+Ask the user:
+  - Is your application primarily reading data (analytics, reports, searches) or writing data (logging, IoT ingestion, frequent updates)?
```
addaleax: This will vary between collections quite a bit, it's normal (and encouraged) to separate write-heavy collections from read-heavy collections

skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,45 @@ Reference these guidelines when:
...
+
+**Step 3: Work with the user to choose the best source(s)**
+  Recommend the best source(s) for their situation, explaining the tradeoffs. For schema design decisions, we often need to combine multiple sources for a complete picture.
+
+**Step 4: Proceed with analysis**
+  Only after source selection, fetch data or guide the user through analysis.
+
+#### Sources
+
+- [Query statistics](references/source-query-stats.md) - Returns runtime statistics for recorded queries showing query shapes and frequency. **Limitation**: Currently only captures `find`, `aggregate` and `distinct` operations - best for read heavy workloads. Requires Atlas M10+ tier.
```
addaleax: "best for read heavy workloads" is semi-accurate here – this can be used perfectly fine with balanced workloads too, but the LLM needs to account for the fact that it cannot see the writes here, only relative impact of the reads

skills/mongodb-schema-design/SKILL.md
paula-stacho: Good point, I'll change the phrasing so that it's more about collections we're focusing on

skills/mongodb-schema-design/SKILL.md
paula-stacho: Makes sense, I'll change the phrasing!

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
```diff
@@ -52,15 +52,13 @@ When category data changes (a rare event), use `updateMany` to update all produc
 
 ## Verify with
 
+####  Find pipelines with $lookup stages 
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md)
+Use codebase if available, ask the user.
+As a last resort, use the system profiler (see ./source-system-profile.md)
```
asya999: Couldn't they use logs?  We really don't want to suggest system.profile collection for anything.  Profile command will increase slow query logging into the log files, so reference that rather than system.profile collection.

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
```diff
@@ -52,15 +52,13 @@ When category data changes (a rare event), use `updateMany` to update all produc
...
-db.setProfilingLevel(1, { slowms: 50 }) // Disable afterwards
-db.system.profile.find({
-  "command.aggregate": { $exists: true },
-  "command.pipeline.$lookup": {
-    $exists: true
-  }
-}).sort({ millis: -1 })
 
 // Check if $lookup foreign fields are indexed
 db.reviews.aggregate([
```
asya999: I'm curious why indexStats and not getIndexes() - doesn't indexstats require higher permissions?

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
asya999: how does the agent know which one that would be?  This is for $lookup?  it seems ambiguous

asya999, line 79: this is a problematic phrasing since SBE explain does not show that statistic inside $lookup stage

skills/mongodb-schema-design/references/antipattern-unnecessary-collections.md
```diff
@@ -80,17 +80,16 @@ for (const d of db.adminCommand({ listDatabases: 1 }).databases) {
...
-  { $sort: { count: -1 } }
-])
 // Collections with similar access patterns should be combined
 ```
 
+### Check if collections are always accessed together.
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md)
+Use codebase if available, ask the user.
+As a last resort, use the system profiler (see ./source-system-profile.md)
```
asya999: again, remove reference to system profile collection - replace with slow query logs if something else is needed here.

skills/mongodb-schema-design/references/antipattern-unnecessary-collections.md
```diff
@@ -80,17 +80,16 @@ for (const d of db.adminCommand({ listDatabases: 1 }).databases) {
...
-// Check if collections are always accessed together
-// If orders always needs customer, items, addresses
-// → they should be embedded
-db.system.profile.aggregate([
-  { $match: { op: "query" } },
-  { $group: { _id: "$ns", count: { $sum: 1 } } },
-  { $sort: { count: -1 } }
-])
 // Collections with similar access patterns should be combined
 ```
```
asya999, line 82: it's not clear to me what this means.   What is "similar access patterns"?  How should they be combined?  And how will this show up in. queryStats?

skills/mongodb-schema-design/references/pattern-approximation.md
```diff
@@ -56,15 +56,15 @@ Use Approximation when staleness is acceptable. Use Computed when exact values a
...
-db.setProfilingLevel(1, { slowms: 0 })
-db.system.profile.find({
-  "command.update": "articles",
-  "command.updates.u.$inc.viewCount": { $exists: true }
-}).count()
-// High count relative to read count suggests approximation would help
+### Check write frequency on counter fields
+
+Use codebase if available, ask the user.
+Consult the documentation to see if $queryStats already supports write commands.
```
asya999: what would they be looking for in the write commands?

skills/mongodb-schema-design/references/pattern-approximation.md
```diff
@@ -56,15 +56,15 @@ Use Approximation when staleness is acceptable. Use Computed when exact values a
...
+As a last resort, use the system profiler (see ./source-system-profile.md)
 
+High count relative to read count suggests approximation would help
```
asya999: Is that the case?   Wouldn't that literally only apply if it's high count of writes to a *specific field*?

skills/mongodb-schema-design/references/pattern-approximation.md
asya999: wouldn't it be better to sort by lastSyncedAt, limit 10 and then calculate staleness?  That way if there is an index on the lastSyncedAt field it will be used - this forces a full collection scan with calculation of every single document with an in-memory sort keeping only highest ten.

skills/mongodb-schema-design/references/pattern-computed.md
```diff
@@ -147,25 +147,20 @@ On-demand materialized views are not automatically refreshed—you control when
...
-  }},
-  { $match: { count: { $gt: 100 } } }  // Repeated 100+ times
-])
-// High count + high avgMs = candidate for computed pattern
-```
+### Find expensive aggregations that should be pre-computed
+
+For Atlas M10+ use slow query logs to find the slowest aggregations (see ./source-slow-query-logs.md). 
+Use codebase if available, ask the user.
+As a last resort, use the system profiler (see ./source-system-profile.md)
```
asya999: we should remove

skills/mongodb-schema-design/references/pattern-computed.md
```diff
@@ -147,25 +147,20 @@ On-demand materialized views are not automatically refreshed—you control when
...
+
+### Check if same aggregation runs repeatedly
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md)
+Use codebase if available, ask the user.
+As a last resort, use the system profiler (see ./source-system-profile.md)
+
+
+High count + high avgMs = candidate for computed pattern
```
asya999: wouldn't the aggregation have to be calculating something for this to be the case?

skills/mongodb-schema-design/references/pattern-extended-reference.md
```diff
@@ -64,26 +64,11 @@ Keep both a bare reference (`customerId`) and an optional cache subdocument (`cu
...
-  { $group: { _id: "$lookup.from", count: { $sum: 1 } } }
-])
-// High count = candidate for extended reference
-```
+Find lookup-heavy aggregations. See how often lookups hit the same collection. High count = candidate for extended reference
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md) and slow query logs (see ./source-slow-query-logs.md)
+Use codebase if available, ask the user.
+As a last resort, use the system profiler (see ./source-system-profile.md)
+
```
asya999: I understand removing the query of system.profile aggregation (good) but shouldn't it be replaced by equivalent explanation of what they should be looking for? In the code base or in slow query logs or in queryStats?

asya999: oh i think I see - that code is in query-stats.md file?

skills/mongodb-schema-design/references/source-query-stats.md
```diff
@@ -0,0 +1,75 @@
...
+
+With mcp-server, use the `mcp__mongodb__aggregateDB` tool with database set to `admin`.
+
+```javascript
+db.getSiblingDB("admin").aggregate([{ $queryStats: {} }])
+```
+
+**Example 1: Find collections frequently queried together with others (embedding candidates)**
+```javascript
+db.getSiblingDB("admin").aggregate([
```
asya999: you are sorting each line in query stats - don't you want to aggregate them into sums for specific collections (with specific $lookups)?   Remember each different agg will be a separate line in query stats.  But logically they all might be combining the same two collections...

skills/mongodb-schema-design/references/source-query-stats.md
```diff
@@ -0,0 +1,75 @@
...
+          "$metrics.execCount"
+        ]
+      }
+    }
+  },
+  { $sort: { execCount: -1 } }
+])
+
+// Check pipeline.$lookup.from for joined collections
+// High execCount + high avgMs = consider embedding instead of $lookup
```
asya999: shouldn't this check be part of the above aggregation?

skills/mongodb-schema-design/references/source-query-stats.md
```diff
@@ -0,0 +1,75 @@
...
+**Example 2: Find top most frequent query shapes (optimize hot paths)**
+```javascript
+db.getSiblingDB("admin").aggregate([
+  { $queryStats: {} },
+  { $sort: { "metrics.execCount": -1 } },
+  { $limit: 10 },
+  {
+    $project: {
+      command: "$key.queryShape.command",
+      collection: "$key.queryShape.cmdNs.coll",
```
asya999, line 107: why are you dropping db from this?  collection names are not unique across databases...

skills/mongodb-schema-design/references/source-query-stats.md
```diff
@@ -0,0 +1,75 @@
...
+        $divide: [
+          { $divide: ["$metrics.totalExecMicros.sum", 1000] },
+          "$metrics.execCount"
+        ]
+      }
+    }
+  }
+])
+
+// High execCount = hot path → design your schema for these queries first
```
asya999, line 120: but they could be already really optimal - wouldn't execution time be relevant here too?

skills/mongodb-schema-design/references/pattern-approximation.md
```diff
@@ -56,15 +56,15 @@ Use Approximation when staleness is acceptable. Use Computed when exact values a
...
-db.system.profile.find({
-  "command.update": "articles",
-  "command.updates.u.$inc.viewCount": { $exists: true }
-}).count()
-// High count relative to read count suggests approximation would help
+### Check write frequency on counter fields
+
+Use codebase if available, ask the user.
+Consult the documentation to see if $queryStats already supports write commands.
+As a last resort, use the system profiler (see ./source-system-profile.md)
```
asya999: again, remove -replace with slow query logs

skills/mongodb-schema-design/references/pattern-computed.md
```diff
@@ -147,25 +147,20 @@ On-demand materialized views are not automatically refreshed—you control when
...
+
+For Atlas M10+ use slow query logs to find the slowest aggregations (see ./source-slow-query-logs.md). 
+Use codebase if available, ask the user.
+As a last resort, use the system profiler (see ./source-system-profile.md)
+
+### Check if same aggregation runs repeatedly
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md)
+Use codebase if available, ask the user.
+As a last resort, use the system profiler (see ./source-system-profile.md)
```
asya999: slow query logs

skills/mongodb-schema-design/references/pattern-extended-reference.md
```diff
@@ -64,26 +64,11 @@ Keep both a bare reference (`customerId`) and an optional cache subdocument (`cu
...
-  { $match: { "lookup": { $exists: true } } },
-  { $group: { _id: "$lookup.from", count: { $sum: 1 } } }
-])
-// High count = candidate for extended reference
-```
+Find lookup-heavy aggregations. See how often lookups hit the same collection. High count = candidate for extended reference
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md) and slow query logs (see ./source-slow-query-logs.md)
+Use codebase if available, ask the user.
+As a last resort, use the system profiler (see ./source-system-profile.md)
```
asya999: remove

skills/mongodb-schema-design/references/source-slow-query-logs.md
```diff
@@ -0,0 +1,61 @@
...
+- Query details (filter, pipeline, etc.)
+- Execution metrics (executionTimeMillis, docsExamined, planSummary, etc.)
+
+## What to Look For
+
+When analyzing slow query logs, focus on:
+
+**Slow $lookup operations:**
+- Look for `$lookup` in the log line
+- Slow $lookup = candidate for embedding or denormalization
```
asya999: what is slow?  Isn't the fact it's already logged mean it's slow?

skills/mongodb-schema-design/references/source-slow-query-logs.md
```diff
@@ -0,0 +1,61 @@
...
+- Cross-reference with Query Stats to identify frequent lookups
+- High executionTimeMillis + high frequency = urgent schema redesign
+
+**Schema design insights:**
+- Queries on related collections (multiple namespaces accessed together) → consider embedding
+- Operations on array fields with high docsExamined → potential unbounded array issues
```
asya999: that's ... not true.  docsExamined is the number of documents.   Unbounded array is a large array *inside* a single document... so not sure what this is meant to look for.

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
+# System Profiler
+
+## When to use
+
+Important: Can significantly degrade production performance. Only use when other sources are unavailable or insufficient. Explain the impact. Use with caution, limit time and scope.
```
asya999: They should never use profile command with levels 1 and 2.   Level 0 with argument to it will change what's logged in the slow query logs and that's acceptable use with caution.

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+
+Collects detailed information about Database Commands executed against a running mongod instance, including CRUD operations and administration commands.
```
asya999: This is incomplete and hence incorrect.   It increases the threshold for both slow query logs *AND* when level is 1 writes operations to system.profile collection (and 2 writes everything into profile collection and hence should never be recommended).

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Requirements
+
+The profiler is off by default. It can be enabled per-database or per-instance at one of several profiling levels. Before enabling, work with the user to determine the profiling level and duration to enable, balancing the need for detailed information with the impact on performance. For most cases, we recommend enabling at level 1 with a reasonable `slowms` threshold (e.g. 50ms), and disabling after the analysis is complete.
```
asya999: no, I'd say we only recommend lower slowMs threshold if more information is needed in slow query logs.

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+
+```javascript
+// enable
+db.setProfilingLevel(1, { slowms: <number> })
+
+// ... do your analysis ...
+
+// disable
+db.setProfilingLevel(0)
```
asya999: THIS DOES NOT SET SLOWMS back to what it was (and this is why we should not ever recommend this)

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+## How to use
+
+Example queries to identify access patterns:
```
asya999: I'm going to make comments on these not because I think we should keep them but because slow query logs would have the same format.

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Example 1: Find pipelines with $lookup stages (embedding candidates)**
+```javascript
+db.setProfilingLevel(1, { slowms: 50 }) // Disable afterwards
```
asya999: Disable and set back to what slowms was before!

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+  "command.pipeline.$lookup": {
+    $exists: true
+  }
+}).sort({ millis: -1 })
+```
+
+**Example 2: Check if collections are always accessed together (embedding candidates)**
+```javascript
+// If orders always needs customer, items, addresses
+// → they should be embedded
```
asya999: I don't see how this aggregation works - isn't it grouping things by namespace and counts how many queries there are on each???  How does that correlate them if they are related or not?

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+])
+```
+
+**Example 3: Check write frequency on counter fields (approximation candidates)**
+```javascript
+db.setProfilingLevel(1, { slowms: 0 })
+db.system.profile.find({
+  "command.update": "articles",
+  "command.updates.u.$inc.viewCount": { $exists: true }
+}).count()
```
asya999: is that meaningful without knowing what time slice is being analyzed?  It's just a single number

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+**Example 4: Find expensive aggregations that should be pre-computed**
+```javascript
+db.setProfilingLevel(1, { slowms: 100 }) // Disable afterwards
+db.system.profile.find({
+  "command.aggregate": { $exists: true },
+  millis: { $gt: 100 }
```
asya999: they are all going to be greater than 100

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+}).sort({ millis: -1 })
```
asya999: without limit this sorts entire profile collection

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+  "command.update": "articles",
+  "command.updates.u.$inc.viewCount": { $exists: true }
+}).count()
+```
+
+**Example 4: Find expensive aggregations that should be pre-computed**
+```javascript
+db.setProfilingLevel(1, { slowms: 100 }) // Disable afterwards
+db.system.profile.find({
+  "command.aggregate": { $exists: true },
```
asya999: we don't know they are doing any computation though

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+  millis: { $gt: 100 }
+}).sort({ millis: -1 })
+```
+
+**Example 5: Check if same aggregation runs repeatedly**
+```javascript
+db.system.profile.aggregate([
+  { $match: { "command.aggregate": { $exists: true } } },
```
asya999: is it intentional that this groups same aggregation even if it ran on many different collections (like you run countDocuments on 100 collections and this will group it and say that you ran countDocuments 100 times)

skills/mongodb-schema-design/references/source-system-profile.md
```diff
@@ -0,0 +1,91 @@
...
+**Example 6: Check how often lookups hit same collections**
+```javascript
+// Check how often lookups hit same collections
+db.system.profile.aggregate([
+  { $match: { "command.pipeline.$lookup": { $exists: true } } },
+  { $project: { pipeline: "$command.pipeline" } },
+  { $unwind: "$pipeline" },
+  { $project: { lookup: { $getField: { field: { $literal: '$lookup' }, input: '$pipeline' } } } },
+  { $match: { "lookup": { $exists: true } } },
+  { $group: { _id: "$lookup.from", count: { $sum: 1 } } }
```
asya999: so these might be different original collections all doing $lookup in this one collection.. should it be embedded into all of them?  Also this is unnecessarily complex.  you don't need all those extra projects just to rename fields...

skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,45 @@ Reference these guidelines when:
...
+Ask the user:
+  - Is this a new design or is there a production database with existing access patterns to analyze?
+  - If there is production data, is it on Atlas? If yes, what tier? (M0/M2/M5 vs M10+)
+
+**Step 2: Determine workload type**
+Is the workload read-heavy, write-heavy, or balanced? This will influence which diagnostic sources are most relevant.
+Ask the user:
+- What's the primary workload for these collections — read-heavy (analytics, reports, searches), write-heavy (logging, IoT ingestion, frequent updates), or balanced?
+
+Verify with `db.serverStatus().opcounters`.
```
asya999, line 70: opcounters is very problematic for this.  It includes aggregations in commands.  also findAndModify is in commands.   It includes system command/queries/operations also and doesn't split things by namespaces.

skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,45 @@ Reference these guidelines when:
...
+
+**Step 3: Work with the user to choose the best source(s)**
+  Recommend the best source(s) for their situation, explaining the tradeoffs. For schema design decisions, we often need to combine multiple sources for a complete picture.
+
+**Step 4: Proceed with analysis**
+  Only after source selection, fetch data or guide the user through analysis.
+
+#### Sources
+
+- [Query statistics](references/source-query-stats.md) - Returns runtime statistics for recorded queries showing query shapes and frequency. **Limitation**: Currently only captures `find`, `aggregate` and `distinct` operations (pair with other sources for write patterns). Requires Atlas M10+ tier.
```
asya999: we keep saying `find` `aggregate` and `distinct` but it captures *all reads* we should just be saying reads (as opposed to writes) not enumerating them like some reads might be missing.

skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,45 @@ Reference these guidelines when:
...
+- [Atlas Slow Query Logs](references/source-slow-query-logs.md) - Review slow queries (actual queries, not shapes) to identify performance bottlenecks. Captures all operation types including writes. Requires Atlas M10+ tier.
```
asya999: say "captures all reads and writes"

skills/mongodb-schema-design/SKILL.md
```diff
@@ -51,6 +51,45 @@ Reference these guidelines when:
...
+- Codebase - Examine actual queries in application code to understand access patterns, especially for new applications or with changing workloads. Can be used in conjunction with query stats for a more complete picture.
+- Natural language input - Ask the user to describe their typical queries and access patterns in natural language. Can be used as the only source or to supplement and validate other sources - the user might have contextual knowledge that is not reflected in the data or codebase.
+- [System Profiler](references/source-system-profiler.md) - Can significantly degrade production performance. Propose as an alternative when other sources are unavailable or insufficient, explain the impact. Collects detailed information about Database Commands executed against a running mongod instance, including CRUD operations and administration commands.
```
asya999: please don't recommend turning on system.profile ever.

testing/mongodb-schema-design/evals/evals.json
```diff
@@ -80,6 +80,19 @@
...
+    {
+      "id": 7,
+      "name": "profiler-vs-atlas-sources",
+      "prompt": "The performance of my Atlas M10 cluster has degraded recently. I want to enable the profiler to see what queries are slow. Can you help me set that up so I can figure out what's wrong with my schema?",
+      "expected_output": "Should recognize the user wants to use System Profiler but explain that Atlas provides better alternatives: Query Stats (for frequent patterns) and Slow Query Logs (for slow operations). Should explain the tradeoffs - profiler has performance overhead and requires enabling/disabling, while Atlas sources are always-on and production-safe. Should recommend using Query Stats + Slow Query Logs together for comprehensive analysis without profiler overhead.",
+      "expectations": [
+        "Recognizes that System Profiler is available but has significant performance overhead in production",
+        "Explains Query Stats and Slow Query Logs as Atlas-specific alternatives that don't degrade performance",
+        "Compares the sources: System Profiler (performance impact, all operations) vs Query Stats (no overhead, frequent patterns) vs Slow Query Logs (actual slow queries)",
```
asya999: add "significant" before "performance impact"

asya999 (commented):
Some corrections, questions, and also would like us to strongly de-emphasize using system.profile ever, and only mention profiling if they want to slower `slowms` for logs.

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
paula-stacho: Removing system profile mention.

skills/mongodb-schema-design/references/antipattern-unnecessary-collections.md
paula-stacho: Removing system profile mention.

skills/mongodb-schema-design/references/pattern-computed.md
paula-stacho: Removing system profile mention.

skills/mongodb-schema-design/references/pattern-approximation.md
paula-stacho: Removing system profile mention.

skills/mongodb-schema-design/references/pattern-extended-reference.md
paula-stacho: Removing system profile mention.

skills/mongodb-schema-design/references/pattern-computed.md
paula-stacho: Removing system profile mention.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/source-system-profile.md
paula-stacho: Removing file.

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
paula-stacho: Good question! From the documentation, it seems the permissions are not exactly higher but different (`clusterMonitor` role would have `indexStats` but `listIndexes` limited to the config database - [source](https://www.mongodb.com/docs/manual/reference/built-in-roles/?deployment-type=self#mongodb-authrole-clusterMonitor)). It's quite possible our user might lack read permission to the production database, but looks like this assumption is not clear/consistent among the skill (e.g. the example right below). For now, I can add the other example, but in general the read vs monitor permissions is something that should be addressed - I'll bring this back to the team.

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
paula-stacho: Not sure if I follow here - which query or which index? This is an example of checking existence of indexes on the foreign keys, after the "problematic" queries have been identified

paula-stacho: I admit the examples are rather crude and depend on the agent already having some knowledge, in my understanding these are just some pointers to "nudge" the agent into the right direction. cc @addaleax

skills/mongodb-schema-design/references/pattern-approximation.md
paula-stacho: This note was added as a replacement for the system.profile approach - now we mostly recommend queryStats but that doesn't include writes at the moment. On second thought, I'll remove this note as we already have this information on the top level.

skills/mongodb-schema-design/references/pattern-approximation.md
paula-stacho: Yes, agreed that this note is less clear now when the context above is gone. Will change the wording.

skills/mongodb-schema-design/SKILL.md
paula-stacho: Updated!

skills/mongodb-schema-design/SKILL.md
paula-stacho: I was following the docs on this (source: https://www.mongodb.com/docs/manual/reference/operator/aggregation/queryStats/#mongodb-pipeline-pipe.-queryStats), but I'll gladly update for simplicity

skills/mongodb-schema-design/SKILL.md
paula-stacho: thank you @asya999 , these are good points especially on the namespaces. do you have an alternative in mind that we could recommend? otherwise I can just remove the note.

skills/mongodb-schema-design/references/source-slow-query-logs.md
paula-stacho: Yes exactly, this wasn't meant as "look for slow ones as a next step", rather "consider embedding". I'll update the wording, let me know if it's clearer!

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
asya999: it says "look for index supporting the query" - I'm not sure how/whether the agent has the knowledge to "figure out" which index(es) can support the $lookup query.

skills/mongodb-schema-design/SKILL.md
asya999: if you run `db.collName.aggregate({$collStats:{latencyStats: {}}})` you see something like this: 
```
[
  {
    ns: 'test.tickets',
    host: 'M-GC9Q974V14:27017',
    localTime: ISODate('2026-06-02T14:19:32.821Z'),
    latencyStats: {
      commands: {
        latency: Long('NumberLong(50120)'),
        ops: Long('NumberLong(2)'),
        queryableEncryptionLatencyMicros: Long('NumberLong(0)')
      },
      reads: {
        latency: Long('NumberLong(169466)'),
        ops: Long('NumberLong(64)'),
        queryableEncryptionLatencyMicros: Long('NumberLong(0)')
      },
      writes: {
        latency: Long('NumberLong(204752)'),
        ops: Long('NumberLong(9003)'),
        queryableEncryptionLatencyMicros: Long('NumberLong(0)')
      },
      transactions: {
        latency: Long('NumberLong(0)'),
        ops: Long('NumberLong(0)'),
        queryableEncryptionLatencyMicros: Long('NumberLong(0)')
      }
    }
  }
]
```
So it gives you commands, reads, writes and transactions for the namespace (the relevant field is `ops` for number of operations).  However that is *not* the number of documents affected so insertMany with 1000 documents is one operation, which is the other issue with ops.  Also findOne and find returning 1000 are both one operation.  And update that matches nothing is still one write.  But the nice thing here is that aggregations are counted as reads and not commands.  So *ballpark* this should maybe work ok.  It would be better even to run it a few minutes apart if this is a live collection and compare rather than looking at absolute numbers (which are since last restart and may not be useful).

skills/mongodb-schema-design/SKILL.md
asya999: You list it as "limitation" where in the docs it's a clarification or enumeration (to confirm it really is all reads).  It's okay to say it but not with "limitation" highlighted note which suggests the opposite of what it should (it suggests that we don't capture all reads but only limited to the ones enumerated).

asya999: In fact, it's also the word "only" that adds to the misleading meaning.

skills/mongodb-schema-design/references/source-slow-query-logs.md
paula-stacho: Thank you Asya, removing!

skills/mongodb-schema-design/references/source-query-stats.md
paula-stacho: the intention here wasn't to say that they necessarily need optimization, more to highlight the hot paths as a crucial factor for the schema design (in line with the "what is accessed together.."). I'll add a note on cross-referencing this with $avgMs to identify areas for improvement.

skills/mongodb-schema-design/references/source-query-stats.md
paula-stacho: thank you for noticing!

skills/mongodb-schema-design/SKILL.md
paula-stacho: thanks for the explanation!

skills/mongodb-schema-design/SKILL.md
paula-stacho: Thank you so much @asya999 , this looks very helpful!

skills/mongodb-schema-design/references/pattern-extended-reference.md
paula-stacho: yes, examples that use a source have been moved to the corresponding "source" files

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
paula-stacho: true. there is more detailed information on this in the query optimization skill, but as we can't cross-reference skills, I'll add a bit more context to the note.

paula-stacho: I have ran a few evals locally with lower models (claude sonnet and haiku) and they did indeed make some mistakes, adding a bit of context like this message helped.

skills/mongodb-schema-design/references/antipattern-unnecessary-collections.md
paula-stacho: I'll remove this note, it's a leftover from the example that used to be there and it seems redundant, there is more detail in the section below

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
paula-stacho: thank you for the info @asya999 ! I'm not sure what would be the best replacement to "Measure $lookup impact", perhaps we can use the time if `executionTimeMillisEstimate` is available in most engines? wdyt @addaleax ?

skills/mongodb-schema-design/references/pattern-approximation.md
paula-stacho: agreed, I'll update the example

skills/mongodb-schema-design/references/pattern-computed.md
paula-stacho: agreed, will change the wording!

skills/mongodb-schema-design/references/source-query-stats.md
paula-stacho: update: I split this into two examples, one is focused on finding the collection combinations, another as a follow up to see more details on the query shapes

skills/mongodb-schema-design/references/source-query-stats.md
paula-stacho: addressing together with the point above

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
asya999: I’m not sure it’s meaningful to sat measure $lookup impact.  Yes a single run of $lookup should be as efficient as possible but how many ops are $lookup is probably even more impactful

skills/mongodb-schema-design/references/pattern-approximation.md
```diff
@@ -56,25 +56,23 @@ Use Approximation when staleness is acceptable. Use Computed when exact values a
...
+### Check write frequency on counter fields
+
+Use codebase if available, ask the user.
+
+High count relative to read count on a specific field suggests approximation would help
 
+```javascript
 // Compare counter staleness
 db.articles.aggregate([
+  { $sort: { staleness: -1 } },
```
asya999: Why did this move?  You can’t sort of a field when it doesn’t exist yet.

paula-stacho:
@dacharyc The validate skill check if failing because of a warning (the tokens in ref files are a bit over the soft limit). I'm not sure we can squeeze under the limit now, but have we considered making it non strict or splitting the checks? It's not a required check but I'd still avoid merging it red if possible, to not hide potential other issues in the future

asya999 (commented):
sorry forgot to hit send on this!

asya999 (commented):
Sort of last synced should be ascending

dacharyc:
> @dacharyc The validate skill check if failing because of a warning (the tokens in ref files are a bit over the soft limit). I'm not sure we can squeeze under the limit now, but have we considered making it non strict or splitting the checks? It's not a required check but I'd still avoid merging it red if possible, to not hide potential other issues in the future

Hey @paula-stacho - I have confirmed with the team that we'll continue to enforce this as a strict check for the moment. I've taken a look around the reference files and think it's probably possible to trim the tokens to get under the limit. Can you make a best effort to see what can be rewritten to be more concise/cut entirely to see if we can get this bundle under the limit?

skills/mongodb-schema-design/references/pattern-approximation.md
```diff
@@ -56,25 +56,23 @@ Use Approximation when staleness is acceptable. Use Computed when exact values a
...
+### Check write frequency on counter fields
+
+Use codebase if available, ask the user.
 
+High count relative to read count on a specific field suggests approximation would help
+
+```javascript
 // Compare counter staleness
 db.articles.aggregate([
+  { $sort: { lastSyncedAt: -1 } },
```
asya999: this looks at most recent synced - wasn't the intention to see oldest last sync'ed?   Used to sort most stale first (ten oldest) now with -1 on lastSync'ed it's ten newest.

asya999 (commented):
i think this comment was lost before

skills/mongodb-schema-design/references/pattern-approximation.md
paula-stacho: yes thank you @asya999 !

paula-stacho:
> > @dacharyc The validate skill check if failing because of a warning (the tokens in ref files are a bit over the soft limit). I'm not sure we can squeeze under the limit now, but have we considered making it non strict or splitting the checks? It's not a required check but I'd still avoid merging it red if possible, to not hide potential other issues in the future
> 
> Hey @paula-stacho - I have confirmed with the team that we'll continue to enforce this as a strict check for the moment. I've taken a look around the reference files and think it's probably possible to trim the tokens to get under the limit. Can you make a best effort to see what can be rewritten to be more concise/cut entirely to see if we can get this bundle under the limit?

Thank you for your reply @dacharyc . I couldn't identify any obvious cuts, definitely not enough to get under the limit. I am also cautious of the amount of changes we're making in this PR and don't want to increase the scope of the changes (and this project) further. 

Would it be possible to slightly increase the soft limit or skip it for this skill, and perhaps we can plan some future changes? For example:
- consider splitting the skill
- make bigger cuts - if we confirm that we expect this particular skill to be used with higher models and so the skill should focus on supplementing their training with the less-documented information, we could cut a lot of the basics

skills/mongodb-schema-design/references/antipattern-excessive-lookups.md
```diff
@@ -52,21 +52,24 @@ When category data changes (a rare event), use `updateMany` to update all produc
 
 ## Verify with
 
+####  Find pipelines with $lookup stages 
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md)
```
dacharyc: It looks like we're using filepaths here relative to the `references` directory. The Agent Skills specification calls for links to other files to be relative to the skill root: https://agentskills.io/specification#file-references

So here and elsewhere where we're referring to other reference files, we should update the links to be relative to skill root - i.e.:


```suggestion
For Atlas M10+ use $queryStats. See [Query Stats](references/source-query-stats.md)
```

skills/mongodb-schema-design/references/pattern-approximation.md
```diff
@@ -56,25 +56,23 @@ Use Approximation when staleness is acceptable. Use Computed when exact values a
...
-db.system.profile.find({
-  "command.update": "articles",
-  "command.updates.u.$inc.viewCount": { $exists: true }
-}).count()
-// High count relative to read count suggests approximation would help
+### Check write frequency on counter fields
+
+Use codebase if available, ask the user.
 
+High count relative to read count on a specific field suggests approximation would help
```
dacharyc, line 63: I should have flagged this in the comment from the prior version - is there any guidance we can provide for what an agent should consider a "high count relative to read count?" Without more specific guidance, different agents may consider different counts "high" here.

skills/mongodb-schema-design/references/pattern-computed.md
```diff
@@ -147,25 +147,16 @@ On-demand materialized views are not automatically refreshed—you control when
...
+
+For Atlas M10+ use slow query logs to find the slowest aggregations (see ./source-slow-query-logs.md). 
+Use codebase if available, ask the user.
+
+### Check if same aggregation runs repeatedly
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md)
+Use codebase if available, ask the user.
+
+High count + high avgMs on an aggregation that computes a result = candidate for computed pattern
```
dacharyc, line 160: Similar question here - is there any guidance we can provide an agent for what it should consider a "high count" or a "high avgMs"?

skills/mongodb-schema-design/references/pattern-extended-reference.md
```diff
@@ -64,26 +64,9 @@ Keep both a bare reference (`customerId`) and an optional cache subdocument (`cu
...
-  { $match: { "command.pipeline.$lookup": { $exists: true } } },
-  { $project: { pipeline: "$command.pipeline" } },
-  { $unwind: "$pipeline" },
-  { $project: { lookup: { $getField: { field: { $literal: '$lookup' }, input: '$pipeline' } } } },
-  { $match: { "lookup": { $exists: true } } },
-  { $group: { _id: "$lookup.from", count: { $sum: 1 } } }
-])
-// High count = candidate for extended reference
-```
+Find lookup-heavy aggregations. See how often lookups hit the same collection. High count = candidate for extended reference
```
dacharyc, line 67: Similar question here re: high count - can we be more specific?

skills/mongodb-schema-design/references/source-query-stats.md
```diff
@@ -0,0 +1,123 @@
...
+          { $divide: ["$metrics.totalExecMicros.sum", 1000] },
+          "$metrics.execCount"
+        ]}
+      }
+    }
+  },
+  { $sort: { totalLookupHits: -1 } }
+])
+
+// High totalLookupHits = frequently joined
```
dacharyc, line 61: Similar questions here re: what we mean by "high" in these comments. More specific guidance will help agents make more consistent decisions.

skills/mongodb-schema-design/references/source-query-stats.md
```diff
@@ -0,0 +1,123 @@
...
+        $divide: [
+          { $divide: ["$metrics.totalExecMicros.sum", 1000] },
+          "$metrics.execCount"
+        ]
+      }
+    }
+  }
+])
+
+// High execCount = hot path → design your schema for these queries first
```
dacharyc, line 120: Similar question here - what would we consider a "high" execCount, and below what would we consider a "slow query"?

Also, re: slow query logs - if we expect an agent to follow the steps in the `source-slow-query-logs.md` file to retrieve log lines for slow queries, we should probably link to it here so an agent knows the resource is available.

skills/mongodb-schema-design/references/source-slow-query-logs.md
```diff
@@ -0,0 +1,61 @@
+# Atlas Slow Query Logs
+
+## When to use
+
+Retrieves log lines for slow queries as determined by the Performance Advisor. Use to identify slow queries and performance bottlenecks. Provides actual query examples (not shapes) with execution times. Captures all operation types including writes, unlike Query Stats which currently only covers find/aggregate/distinct.
+
+## Requirements
+
+- Atlas M10+ cluster
+- Atlas API credentials configured
```
dacharyc, line 10: Regarding API credentials - is there some guidance we may want to provide for an agent to determine whether Atlas API credentials are configured? Is there a health or ping endpoint we might want to provide that an agent can use to validate this before proceeding?

skills/mongodb-schema-design/references/source-slow-query-logs.md
```diff
@@ -0,0 +1,61 @@
...
+- Timestamp (beginning of line)
+- Operation type (command: find, aggregate, update, etc.)
+- Query details (filter, pipeline, etc.)
+- Execution metrics (executionTimeMillis, docsExamined, planSummary, etc.)
+
+## What to Look For
+
+When analyzing slow query logs, focus on:
+
+**Slow $lookup operations:**
```
dacharyc, line 55: Is there any guidance we might want to set for what an agent should consider "slow"?

skills/mongodb-schema-design/references/source-slow-query-logs.md
```diff
@@ -0,0 +1,61 @@
...
+- Look for `$lookup` in the log line
+- Consider embedding to reduce slow $lookup operations
+- Cross-reference with Query Stats to identify frequent lookups
+- High executionTimeMillis + high frequency = urgent schema redesign
```
dacharyc, line 59: Is there any guidance we might want to provide here for what an agent should consider a "high" `executionTimeMillis` or frequency?

dacharyc (commented):
Hey @paula-stacho - there are a few places here where we should change references to other files to be relative to the skill root, per the Agent Skills specification. Also, I've got some Qs throughout about whether we can provide more specific guidance for agents about what constitutes "high" or "slow" - but I recognize those might be harder to pin down/relative, so those aren't blockers to merge.

skills/mongodb-schema-design/references/source-query-stats.md
paula-stacho: Adding the link

skills/mongodb-schema-design/references/source-slow-query-logs.md
paula-stacho: There is no health endpoint, and it would have to be an authenticated endpoint to validate the credentials. I'm adding link to the docs to this section to help with troubleshooting, but the main aim here is to gauge the viability of this option.

paula-stacho:
Thank you @dacharyc ! Updated the link. As for the guidance being a bit vague and relative at times - I think this is actually in line with our intended use of the skill. The agent should act as an advisor, explaining the relevant factors. It can propose a solution, but the final decision should be on the user -  precisely because some factors are relative and context-dependent. 

>  "**Your database, your decision.** I'm here to advise, not to act unilaterally.".

skills/mongodb-schema-design/references/antipattern-unnecessary-collections.md
```diff
@@ -79,18 +79,13 @@ for (const d of db.adminCommand({ listDatabases: 1 }).databases) {
...
-  { $match: { op: "query" } },
-  { $group: { _id: "$ns", count: { $sum: 1 } } },
-  { $sort: { count: -1 } }
-])
-// Collections with similar access patterns should be combined
 ```
 
+### Check if collections are always accessed together.
+
+For Atlas M10+ use $queryStats (see ./source-query-stats.md)
```
dacharyc: This link should be updated to be relative to the skill bundle root.

dacharyc (approved):
There's one more link that needs to be updated, but after that link, LGTM on the structural quality.

Regarding this comment, though:

> As for the guidance being a bit vague and relative at times - I think this is actually in line with our intended use of the skill. The agent should act as an advisor, explaining the relevant factors. It can propose a solution, but the final decision should be on the user - precisely because some factors are relative and context-dependent.

The agent acting as an advisor is not mutually exclusive with providing the agent more guidance about _how_ to evaluate and flag things to the user. In fact, providing clearer guidance would make the agent a more consistent advisor. To pick one example:

> High count + high avgMs on an aggregation that computes a result = candidate for computed pattern

On one run through, an agent may consider, for example, 96 a high count, and 126 a high avgMs, and recommend a computed pattern. On a second run through of the same skill, the agent may _not_ consider this a high count or a high avgMs and may _not_ recommend a computed pattern. Because we don't provide specific details, we're asking every agent to make a subjective call every time it invokes the skill. Same data, different decision and outcome.

I'm not going to block merging on this because I know this has been in progress for a long time and we want to release the updates. But for future updates to this skill, we should aim to provide more clarity in instructional guidance. This is something our QA reviews will flag, and we may add automated tooling around at some point, so I'm happy to answer any questions you might have about what this could look like in practice or why it's important.

skills/mongodb-schema-design/references/antipattern-unnecessary-collections.md
paula-stacho: Ah thanks! I forgot to push that change

PR #40, opened 2026-05-07
chore: add release docs

nirinchev:
Adds CONTRIBUTING.md which focuses on the release process for the different plugins.

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -7,6 +7,7 @@ on:
     paths:
       - ".cursor-plugin/plugin.json"
       - ".claude-plugin/plugin.json"
+      - ".codex-plugin/plugin.json"
       - "gemini-extension.json"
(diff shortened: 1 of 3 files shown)


dacharyc (approved):
This looks fine overall, but I'm wondering if we might want to add a table here that shows when each plugin's release is "released?"

i.e. for Codex, Gemini, GitHub Copilot, looks like it's "released" on "Merge to Main", but for Claude, Cursor, "released" is after "plugin publish." I think a more scannable format might make it easy/fast to understand.

Second, should we mention that the VS Code Extension also requires a coordinated release to pick up Skill changes?

nirinchev:
@dacharyc updated the docs with your feedback - LMK if that's what you had in mind.

dacharyc (approved):
Thanks for adding the details about the VS Code release process and the table - that's exactly what I had in mind. 

LGTM ✅

PR #51, opened 2026-06-26
Add skill-gate CI workflows

dacharyc:
Jira: https://jira.mongodb.org/browse/DOCSP-61223

This PR adds two workflows:
- `skill-gate.yml`: runs the `skill-gate` tool to check skill content for patterns that warrant a security review: https://github.com/mongodb/skill-gate
- `skill-gate-required.yml`: a UX enhancement to provide a "pending" status check as a default state when the `skill-gate` tool hasn't run yet, and resets to `pending` if a new commit pushes after `skill-gate` has run

## Secret storage

The Grove API key is stored as a **GitHub Actions environment secret** in the `skill-gate` environment (gated behind optional required reviewers). The two-job split is preserved: untrusted PR content is checked out in the unprivileged `collect` job, while the key lives only in the privileged `scan` job — they never share a runner.

### To configure before merging

**`skill-gate` environment secret:**
- `GROVE_API_KEY` — the Grove API key

**Repository variable:**
- `GROVE_BASE_URL` — Grove base URL (non-sensitive)

**Branch ruleset (`main`):**
- Set `skill-gate` as a required status check

These workflows are currently in draft pending security review. After approval, add the secret/var above and set `skill-gate` as a required check in the `main` branch ruleset before merging.


.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,237 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Written to RUNNER_TEMP so the data-only checkout's clean step (below)
+          # doesn't wipe it from the workspace; copied back in after checkout.
+          gh api "repos/${{ github.repository }}/pulls/${{ github.event.issue.number }}/files" --paginate \
+            --jq '.[].filename' \
+            | grep '^skills/' | cut -d/ -f1-2 | sort -u > "$RUNNER_TEMP/changed_skills.txt" || true
+          count=$(grep -c . "$RUNNER_TEMP/changed_skills.txt" || true)
+          echo "count=${count:-0}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
cbullinger: Agree with this finding — it's a fail-open, and it's the most consequential of the three because `skill-gate` is intended to become a **required** check.

The `|| true` on the `gh api … | grep '^skills/' | cut -d/ -f1-2 | sort -u > …` pipeline means an API error (rate limit, 5xx) exits 0, `changed_skills.txt` is empty, `count=0`. Then:
- `collect`'s checkout + upload steps gate on `if: steps.changed.outputs.count != '0'` → skipped.
- `scan`'s scan step gates on `if: needs.collect.outputs.count != '0'` → skipped.
- `scan`'s `Report verdict` runs `if: always()` → `verdict_code` is empty → case `""|"0"` → `state=success` ("AUTO-PASS").

Net: an API failure posts AUTO-PASS and lets the PR merge with **no scan** — exactly when the required check should fail closed.

Suggested fix:
```sh
gh api "repos/${{ github.repository }}/pulls/${{ github.event.issue.number }}/files" --paginate \
  --jq '.[].filename' \
  | grep -E '^skills/[^/]+/' | cut -d/ -f1-2 | sort -u > "$RUNNER_TEMP/changed_skills.txt"
count=$(grep -c . "$RUNNER_TEMP/changed_skills.txt" || true)
```
Drop `|| true` from the `gh api` pipeline (let an API failure fail `collect` closed), and match bundle dirs with `^skills/[^/]+/` so non-bundle files like `skills/OWNERS.yaml` don't enter the list. The `|| true` on the `grep -c` is fine to keep (count=0 is a legitimate "no skill changes" outcome, not an error).

.github/workflows/skill-gate-status.yml
```diff
@@ -0,0 +1,50 @@
...
+    # required check stays unseeded → blocked by absence (see NOTE above).
+    if: github.event.pull_request.head.repo.fork == false
+    runs-on: ubuntu-latest
+    steps:
+      - name: Seed required check
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          touched=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" --paginate \
+            --jq '.[].filename' | grep -c '^skills/' || true)
```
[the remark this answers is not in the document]
cbullinger: Agree on the fail-open — same shape as `skill-gate.yml`'s `collect`.

`gh api … --jq '.[].filename' | grep -c '^skills/' || true` — an API error yields `touched=0`, which hits the `if [ "${touched:-0}" -eq 0 ]` branch and seeds `skill-gate=success` ("No skill changes — skill-gate not required"). That lets the PR merge with no scan, which is the wrong default when the API itself failed (vs. legitimately no skill changes).

Suggested fix:
```sh
touched=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
  --paginate --jq '.[].filename' | grep -c '^skills/')
```
Drop `|| true` so an API failure fails the seeder closed instead of silently seeding success.

On the the "only count bundle paths" half: lower priority **here** than in `skill-gate.yml`. In the seeder, over-inclusion just seeds `pending` (which safely blocks merge), so matching broadly is the conservative direction. The fail-open is the part that matters. (Aligning the seeder's `skills/` detection with the scan job's bundle-dir input is still a reasonable cleanliness follow-up, just not load-bearing for safety.)

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,237 @@
...
+            fi
+            echo "## \`$dir\`" >> "$GITHUB_STEP_SUMMARY"
+            set +e
+            # --emit-annotations writes ::error/::warning to stdout (GitHub pins
+            # these to the diff in the PR Checks tab) alongside the markdown report.
+            out=$(skill-gate scan "./$dir/" -o markdown --emit-annotations 2>&1)
+            rc=$?
+            set -e
+            printf '%s\n' "$out"                                             # annotations -> run log / PR Checks
+            printf '%s\n' "$out" | grep -v '^::' >> "$GITHUB_STEP_SUMMARY"   # report -> job summary
```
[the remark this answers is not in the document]
cbullinger: Agree. After `set -e` is re-enabled (the `set +e` … `set -e` only wraps the `skill-gate` invocation and `rc` capture), `grep -v '^::'` exits 1 when **every** line of `$out` is an annotation (starts with `::`) or when `$out` is empty. That fatals the scan step even though `skill-gate` itself succeeded and `rc` is correctly mapped to a verdict.

Suggested fix — make the summary filtering non-fatal:
```sh
printf '%s\n' "$out"                                             # annotations -> run log / PR Checks
printf '%s\n' "$out" | grep -v '^::' >> "$GITHUB_STEP_SUMMARY" || true   # report -> job summary
```
The `grep -v` is cosmetic (job-summary cosmetics), so it must never fail the step on a successful scan. (The first `printf` is unconditional and already safe.)

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,243 @@
...
+          set -euo pipefail
+          # Written to RUNNER_TEMP so the data-only checkout's clean step (below)
+          # doesn't wipe it from the workspace; copied back in after checkout.
+          # Fetch the PR file list on its own line so a `gh api` failure (rate
+          # limit, 5xx) fails this step closed — instead of silently producing
+          # count=0, which skips the scan and posts AUTO-PASS. The grep for
+          # bundle dirs is separate: no match there is a legitimate "no skill
+          # changes" outcome (count=0), not an error, so it stays non-fatal.
+          gh api "repos/${{ github.repository }}/pulls/${{ github.event.issue.number }}/files" --paginate \
+            --jq '.[].filename' > "$RUNNER_TEMP/pr_files.txt"
```
krollins-mdb, line 102: We have a baked-in assumption here that could theoretically allow for bypassing the gate. This GitHub API endpoint is capped at 3,000 changed files. Probably not a number we're realistically going to hit. We have bigger problems with the PR if it's touching that many files.

But maybe worth noting somewhere and codifying PR size expectations in `CONTRIBUTING.md` or something?

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,243 @@
...
+  issue_comment:
+    types: [created]
+
+# Built-in GITHUB_TOKEN only (ephemeral, repo-scoped) — no PAT, no stored token.
+# Permissions are declared per job (least privilege); nothing is granted here.
+permissions: {}
+
+concurrency:
+  group: skill-gate-${{ github.event.issue.number }}
+  cancel-in-progress: true
```
krollins-mdb: I think we may need a stricter concurrency check. The way this is set up now, I believe that after a scan is kicked off, any other unrelated comment can cancel the scan. Example workflow:

1. A maintainer comments `/skill-gate`.
2. This starts run A. It gets concurrency group `skill-gate-51`.
3. Run A passes the job condition, collects the skill data, receives environment approval if needed, and begins scanning.
4. While that scan is still running, anyone comments something unrelated on the PR, such as “Thanks” or “Please address the review.”
5. That comment starts run B, because every `issue_comment` event is a trigger.
6. Before run B reaches the job condition, GitHub sees it belongs to `skill-gate-51` and, because `cancel-in-progress`: true, cancels run A.
7. Run B then evaluates the collect job condition. Its comment was not `/skill-gate`, so collect is skipped. The dependent scan job is skipped too.
8. No scan result is posted by run B. The skill-gate status remains pending (or the successful scan was prevented from completing), and a maintainer must run `/skill-gate` again.

.github/workflows/skill-gate-status.yml
```diff
@@ -0,0 +1,57 @@
...
+# Every push (synchronize) re-seeds against the new head sha, so a stale pass on
+# an old commit never satisfies the new head.
+#
+# NOTE (forks): on fork PRs the built-in GITHUB_TOKEN is READ-ONLY, so the
+# statuses POST would be denied. The job-level `if:` below SKIPS fork PRs to
+# avoid a cosmetic errored run. This is NOT a hole: the `skill-gate` required
+# check is then simply unseeded → stays "Expected" → merge blocked by absence,
+# and the ruleset enforces the check independently of this seeder. Fork PRs here
+# are expected to touch skills/** (new/updated skills), so they need a
+# /skill-gate scan regardless.
```
krollins-mdb: Our handling of forks is odd. I think I remember talking with Dachary about not supporting forks for the workflow. If that seems the most sensible route, we should update this comment and `CONTRIBUTING.md` to indicate we expect a branch workflow, not fork workflow.

krollins-mdb (changes requested):
LGTM, with a couple of things that are more policy related and one thing that's related to workflow functionality.

Requesting changes specifically for the comment workflow, as right now we could potentially waste compute.

.github/workflows/skill-gate-status.yml
cbullinger: I'd prefer to not work with forks; it introduces a lot of complications. I'll update `CONTRIBUTING.md`

.github/workflows/skill-gate.yml
cbullinger: Eek! Good call! Updated

krollins-mdb (approved):
LGTM. Thank you, @cbullinger!

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,251 @@
...
+          name: skill-gate-input-${{ needs.collect.outputs.head_sha }}
+
+      - name: Configure AWS credentials (OIDC)
+        # No static AWS keys in GitHub: the job exchanges GitHub's OIDC token
+        # for short-lived STS creds that may read one secret in the docs AWS
+        # account. The role trust is scoped to
+        # repo:mongodb/agent-skills:environment:skill-gate, so only this
+        # environment can assume it. Actions are SHA-pinned because this is a
+        # public repo and the job holds id-token: write.
+        uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6.2.3
```
nirinchev: Why are we using AWS secret manager instead of storing the secret in the github environment?

cbullinger: Updated to use GH secret

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,226 @@
...
+          echo "verdict_code=$worst" >> "$GITHUB_OUTPUT"
+
+      - name: Report verdict (commit status)
+        if: always()
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          code="${{ steps.scan.outputs.verdict_code }}"
+          case "$code" in
+            ""|"0") state=success; label="AUTO-PASS — no criteria triggered" ;;
```
nirinchev: This reports success for a failed step (the empty case) - as far as I can tell, if something crashes/fails in the `scan` step, we never set `verdict_code`, which means we'll get an empty status here.

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,226 @@
...
+
+  scan:
+    needs: collect
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      statuses: write        # post the `skill-gate` commit status
+    # Gated environment that holds the Grove API key as an environment secret
+    # (GROVE_API_KEY) behind optional required reviewers. Base URL is a
+    # non-sensitive repo var (GROVE_BASE_URL). See setup doc step 3.
```
nirinchev: Should we link the setup doc here?

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,226 @@
...
+# Built-in GITHUB_TOKEN only (ephemeral, repo-scoped) — no PAT, no stored token.
+# Permissions are declared per job (least privilege); nothing is granted here.
+permissions: {}
+
+# Serialize per PR, but do NOT cancel an in-flight run: the /skill-gate filter is a
+# job-level `if:`, so EVERY issue_comment creates a run in this group — with
+# cancel-in-progress, an unrelated comment would kill a scan mid-flight and leave the
+# required check pending. Serializing is safe: the verdict posts to the head SHA resolved
+# at trigger time, and skill-gate-status.yml re-seeds pending if the PR has moved on.
+concurrency:
```
nirinchev, line 48: Not sure I fully understand the concurrency argument here, but do we need the same concurrency rules for skill-gate-status.yml?

.github/workflows/skill-gate.yml
nirinchev: Note that we probably have a separate case for `""` - if `needs.collect.outputs.count` is 0, that should be success, otherwise - failure.

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,226 @@
...
+    # contributor (CONTRIBUTOR/NONE) can't — a member runs it for them.
+    # COLLABORATOR is intentionally omitted: this repo has no outside
+    # collaborators and all contributors are org members.
+    if: >-
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/skill-gate') &&
+      contains(fromJSON('["OWNER","MEMBER"]'), github.event.comment.author_association)
+    runs-on: ubuntu-latest
+    # No `environment:` — this job cannot reach the `skill-gate` environment
+    # secrets (the Grove key), which is the point of the split.
```
nirinchev: This is no longer true and with the rework to use gha secrets, the split probably doesn't make much sense anymore. The only mutable state the github token has access to now is `statuses`, which is a fairly minimal blast radius compare to the `contents: write` it had before we moved off AWS. Should we merge `collect` and `scan` into a single job?

nirinchev (commented):
A bit more thorough review this time - overall, looks good, but there are some corner cases worth looking into and then there are potential further simplifications.

.github/workflows/skill-gate.yml
cbullinger: removed -- that referred to an internal doc

.github/workflows/skill-gate.yml
cbullinger: Added

.github/workflows/skill-gate.yml
cbullinger: done - `collect` and `scan` are a single job. fyi - this will reintroduce the CodeQL alert that the split was avoiding.

.github/workflows/skill-gate.yml
cbullinger: good catch; fixed. the legit empty case and the crashed-scan case are now distinguished

.github/workflows/skill-gate.yml
```diff
@@ -0,0 +1,206 @@
...
+        # pinned skill-gate binary below. See the header comment for why the
+        # CodeQL "untrusted checkout" alert is accepted for this checkout.
+        if: steps.changed.outputs.count != '0'
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          ref: ${{ steps.pr.outputs.head_sha }}
+          persist-credentials: false
+          sparse-checkout: skills
+
+      - name: Stage scan input
```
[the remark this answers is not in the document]
cbullinger: Dismissed as **won't fix** in the code-scanning Security tab. Justification: this is a data-only sparse `skills/` checkout (SHA-pinned to the PR head, `persist-credentials: false`); nothing from the PR is built or executed — only the pinned `skill-gate` binary reads the markdown. The `GITHUB_TOKEN`'s only write scope is `statuses: write` (no `contents: write`), and the Grove key only egresses to Grove behind the `skill-gate` environment's required reviewers. The prior two-job split existed solely to avoid this alert and was collapsed per review. Full write-up in the `skill-gate.yml` header comment.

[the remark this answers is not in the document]
cbullinger: Dismissed as **won't fix** in the code-scanning Security tab. Justification: this is a data-only sparse `skills/` checkout (SHA-pinned to the PR head, `persist-credentials: false`); nothing from the PR is built or executed — only the pinned `skill-gate` binary reads the markdown. The `GITHUB_TOKEN`'s only write scope is `statuses: write` (no `contents: write`), and the Grove key only egresses to Grove behind the `skill-gate` environment's required reviewers. The prior two-job split existed solely to avoid this alert and was collapsed per review. Full write-up in the `skill-gate.yml` header comment.

PR #55, opened 2026-07-23
proposal: functional-grading fields + schema for testing/*/evals.json (draft)

cbullinger:
## The short version

Today, when we test a skill, an AI reads the agent's answer and judges whether it looks right. This PR adds the option to **check whether the answer actually works**: run the query the agent wrote against a real MongoDB with known data in it, and see if it returns what it should.

Nothing here changes existing tests. Every new field is optional, and all **102 test cases across the 7 skills still pass unchanged**. The runner lives in the private `10gen/agent-skills-evals` repo (PR #18). This PR is the public half: the file format, the sample data, and the CI.

## What's in it

**The format + validator.** `testing/evals.schema.json`, enforced by `validate-eval-cases.yml`. Three new optional fields let a test case say "load this data first", "re-run the agent's query and compare results", and "confirm the agent used the right tools". Typos fail in the PR, not mysteriously later.

**Sample data.** Five small scripts that seed known data (movies, orders, events) into a throwaway database, stored next to the test cases that use them.

**Two checks that keep tests honest.**
- `lint-item-echo.mjs` flags tests that copy their answer out of the skill under test (they pass forever and detect nothing). Advisory: comments, never blocks.
- `mutation.yaml` for two skills describes how to deliberately break them, the only way to know "would these tests notice if the skill were wrong?"

**The PR gate.** A maintainer comments `/qa-eval` on a PR. `qa-eval.yml` runs the eval harness against the PR's actual content and posts the verdict as a `qa-eval` check run (report inline in the Checks tab, JSON as artifacts; nothing is posted to the PR conversation). `qa-eval-status.yml` seeds that check on every PR so it can't be skipped. Security shape mirrors skill-gate (#51): `qa-eval.yml` is a single job that does the data-only PR checkout and holds the credentials, the token's only write scope is `checks: write` (no `contents: write`, no `pull-requests: write`), and the Grove key is gated behind the `qa-eval` environment's required reviewers. The CodeQL untrusted-checkout alert is accepted with a header justification, as on #51. Which skills to eval is resolved by `.github/scripts/qa-resolve-skills.sh`, fetched from the default branch at runtime. The data-only checkout deliberately excludes `.github/`, so no PR-provided executable ever runs in the credential-bearing job.

**Two manual workflows for depth.** `qa-pr-screen` (advisory per-PR screen) and `qa-mutation-battery` (runs the mutation configs, publishes sensitivity baselines). `workflow_dispatch` only for now, with no triggers wired yet. `qa-pr-screen.yml`'s header documents what must be restored (the collect/run credential-custody split) before any `pull_request` trigger is enabled.

**Credential failures fail loud in one place.** `harness.py` (in agent-skills-evals#18) names the missing Grove variables at argument time, so these workflows carry no per-repo preflight guards.

## Secret storage

All credentials are GitHub Actions secrets and variables:
- **Grove key**: `GROVE_API_KEY`, a `qa-eval` **environment** secret (gated behind required reviewers, so model spend is deliberate)
- **App PEM**: `AGENT_SKILLS_EVALS_APP_PRIVATE_KEY`, a repo-level secret (needed by both the `qa-eval` and `qa-eval-publish` environments)
- **App ID**: `AGENT_SKILLS_EVALS_APP_ID`, a repo secret (non-sensitive)
- **Grove base URL**: `GROVE_BASE_URL`, a repo variable (non-sensitive)

## Before merging

1. **Everything that spends money ships inert.** The gate workflows sit behind a repo variable (`QA_EVAL_ENABLED`) that isn't set. Screen and battery fire manually only. Enabling sequence: merge this + agent-skills-evals#18, smoke-test via dispatch, then flip the switch.
2. **Don't make `qa-eval` a required check yet.** Not because of forks (check runs work fine on fork PRs) but because the gate has never run in CI. Smoke test first.
3. **Merge this before agent-skills-evals#18.** It has a test that starts enforcing the sample data files once they exist here.

## Why bother

An AI judging an answer can be fooled by a confident-sounding wrong answer, and a test that can't fail isn't evidence of anything. Both have already happened on this corpus. This is the groundwork for saying "this skill works" and showing why.

Coverage is thin by design (1 of 7 skills has a functional test, 2 of 7 a mutation config). Plumbing first: the tests can't be validated until the runner lands.

## Where to look

`testing/evals.schema.json` is the contract. `.github/workflows/qa-eval.yml`'s header documents the gate design and its prereqs. `testing/lint-item-echo.mjs`'s header explains what makes a test case bad, the most transferable idea here.


testing/validate-evals.mjs
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+/**
+ * Validate every testing/<skill>/evals/evals.json against testing/evals.schema.json.
+ * Fails (exit 1) on the first schema violation, printing the skill + JSON path — so a
+ * malformed eval case fails in CI where the builder authors it.
+ *
+ * Run: `npm --prefix testing install && node testing/validate-evals.mjs` (see
+ * .github/workflows/validate-eval-cases.yml). CWD-independent: paths resolve relative to
+ * this file, so it works whether invoked from the repo root or testing/.
```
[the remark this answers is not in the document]
cbullinger: Fixed in 2f29390. All three were accurate: the script reports every violation then exits 1 (not first-failure), prints the file path plus ajv's `instancePath` (not the skill name), and the documented install command now matches CI's `npm ci`. That last one matters beyond tidiness: a local command that resolves different dependency versions than CI can disagree with CI.

I kept report-everything as deliberate rather than changing the code, so an author fixes the whole set in one pass instead of rediscovering one problem per run. Verified by breaking two case files: two failures reported, exit 1.

.github/workflows/validate-eval-cases.yml
```diff
@@ -0,0 +1,29 @@
...
+# malformed eval case (incl. the optional functional_check / trajectory_checks fields) fails
+# where it's authored, not later in the QA harness. Scoped to testing/** so it only runs when
+# eval cases or the schema change.
+on:
+  pull_request:
+    paths:
+      - "testing/**/evals/evals.json"
+      - "testing/evals.schema.json"
+      - "testing/validate-evals.mjs"
+      - "testing/package.json"
```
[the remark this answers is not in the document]
cbullinger: Fixed in 2f29390. Real gap, not cosmetic: `npm ci` installs from the lockfile, so a lockfile-only change (dependency bump, regen) alters what this workflow runs while leaving the workflow unrun. An ajv upgrade that broke validation would have sailed past the check meant to catch it.

`testing/package-lock.json` added to the paths filter.

testing/mongodb-query-optimizer/evals/fixtures/seeded_events.js
```diff
@@ -0,0 +1,24 @@
+// Seed fixture: `seeded_events` — sample.events with NO secondary index (query-optimizer
+// functional case 37). Equality lookups on `userId` do a COLLSCAN until the agent creates
+// an index; scorers.functional_db (type: index) then checks the index exists.
+// Applied per-sample by solvers/seed_db.py.
```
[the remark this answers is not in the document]
cbullinger: Fixed in 2f29390. The header described an earlier design of the case: "functional case 37", equality lookups on `userId`, and an index-existence scorer. The actual consumer is case 20, whose `functional_check` is `type: explain`: filter `{type: 'purchase'}` with sort `{ts: -1}`, requiring an IXSCAN, forbidding an in-memory sort, and capping docsExamined at 1.1x returned.

So what's graded is whether the agent builds a compound index serving both the equality filter and the sort, not merely whether some index exists. Header rewritten to say that, including the line-18 comment flagged as the same problem.

testing/validate-evals.mjs
```diff
@@ -0,0 +1,97 @@
...
+// (build_prompt) or seed_db.py would only discover it deep inside a harness run, as a
+// FileNotFoundError instead of a one-line message pointing at the bad path in this PR.
+function checkAssetsExist(evalsDir, doc) {
+  const problems = [];
+  for (const ev of doc.evals ?? []) {
+    for (const ref of ev.files ?? []) {
+      if (!existsSync(join(evalsDir, ref))) {
+        problems.push(`case ${ev.id}: files entry not found: ${join(evalsDir, ref)}`);
+      }
+    }
```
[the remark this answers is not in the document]
cbullinger: Fixed in 4329d91. The consequence is eval integrity, not filesystem safety, and it's exactly the interaction flagged here: the harness inlines `files` into the prompt, and `lint-item-echo.mjs` deliberately excludes `files` from its overlap analysis because they're provided input data. So `files: ["../../../skills/<name>/SKILL.md"]` hands the agent its own answer key, passes this validator (the file really does exist), and is invisible to the echo lint, the exact teaching-to-the-test failure that lint exists to catch, routed around it.

Fixed in three layers, each catching what the previous can't:

1. **Schema pattern** on `files` items rejecting a leading `/` and any `..` segment. (`seed`'s `^[a-z0-9_]+$` already made traversal unexpressible, which is why only `files` was ever affected.)
2. **`resolve()` + `relative()` containment** in the validator. Switching from `join()` also fixed a real disagreement with the harness: `join('/a','/etc/x')` nests to `/a/etc/x`, while Python's `Path('/a') / '/etc/x'` honours the absolute and yields `/etc/x`, so the two implementations were checking different paths.
3. **`realpathSync` containment** after the existence check. A symlink inside `evals/` needs no `..` and no leading `/`, so it satisfies both checks above. Verified: a `leak.md -> ../../../skills/<name>/SKILL.md` symlink passed cleanly until this branch existed.

Each rejection verified individually (traversal, absolute, mid-path `assets/../../../skills/...`, symlink), a legitimate relative asset still falls through to the ordinary not-found error, and the 102-case corpus still validates clean.

.github/workflows/qa-eval.yml
```diff
@@ -0,0 +1,401 @@
...
+            FAILS=$(jq '[.[] | select(.verdict == "fail")] | length' <<<"$INDEX")
+            INCONCLUSIVE=$(jq '[.[] | select(.verdict == "inconclusive")] | length' <<<"$INDEX")
+            if [ "$FAILS" -gt 0 ]; then
+              CONCLUSION=failure
+              TITLE="QA eval failed"
+            elif [ "$INCONCLUSIVE" -gt 0 ]; then
+              # INCONCLUSIVE = instrument fault (Grove 401, MCP registration failure,
+              # harness crash), not a verdict on the PR. `neutral` is deliberate: it must
+              # not count against the author, and the run should be re-triggered.
+              CONCLUSION=neutral
```
krollins-mdb: I'm curious about why we're trying to avoid counting this "against the author". Using neutral here means that instrument failures can effectively bypass the gate, right? Which means untested stuff could get merged because these failures are considered "neutral", which doesn't block merges.

.github/workflows/qa-eval.yml
```diff
@@ -0,0 +1,401 @@
...
+      - name: Fetch the skill-resolution script (default branch)
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          gh api "repos/${{ github.repository }}/contents/.github/scripts/qa-resolve-skills.sh" \
+            -H "Accept: application/vnd.github.raw" > "$RUNNER_TEMP/qa-resolve-skills.sh"
+          chmod +x "$RUNNER_TEMP/qa-resolve-skills.sh"
+
+      - name: Determine which skill(s) to eval
```
krollins-mdb, line 173: I think the resolver needs a bit more work. It currently only verifies that the requested skill exists. It does not verify that the skill changed in the PR or that all affected skills were evaluated.

This means a PR that touches several skills can receive a successful required check after evaluating only one skill, even an unchanged skill.

.github/workflows/qa-eval-status.yml
```diff
@@ -0,0 +1,84 @@
...
+          # Fetch the file list on its own line so a `gh api` failure fails this seeder
+          # closed -- instead of silently seeding qa-eval=success ("no skill changes")
+          # and allowing merge with no eval. The grep for skill dirs is separate: no
+          # match is a legitimate "no skill changes" outcome (touched=0), not an error.
+          # Known ceiling: this endpoint returns at most 3,000 changed files per PR;
+          # beyond it a skill change could be missed and "success" seeded without an
+          # eval — PR-size expectations live in CONTRIBUTING.md ("Pull requests").
+          files=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
+            --paginate --jq '.[].filename')
+          touched=$(printf '%s\n' "$files" | grep -cE '^skills/[^/]+/' || true)
```
krollins-mdb: Hmm. The seeder and resolver seem to have different definitions of when an eval is required.

- The seeder says testing-only changes do not require evaluation (the line I'm commenting on).
- The runtime resolver says testing-only changes do affect the corresponding skill.

```
skills/<skill>/**
testing/<skill>/**
```

This means a PR can modify the tests, fixtures, functional checks, or mutation definitions that determine evaluation behavior and still receive an immediate successful qa-eval check without running the modified suite.

testing/evals.schema.json
```diff
@@ -0,0 +1,412 @@
...
+        "^_": {
+          "description": "Author comment key (e.g. _note); ignored by the harness."
+        }
+      },
+      "description": "One test case. Single-turn cases use `prompt`; multi-turn cases use `workflow`.",
+      "properties": {
+        "id": {
+          "type": "integer",
+          "description": "Stable case id (unique within the skill; used to correlate results across runs)."
+        },
```
krollins-mdb, line 51: This describes each case ID as unique within its skill. But ]neither the schema nor `validate-evals.mjs` enforces that uniqueness. Duplicate IDs can therefore pass validation and cause unrelated cases to be combined or misidentified during cross-run and mutation analysis.

.github/workflows/qa-eval.yml
```diff
@@ -0,0 +1,401 @@
...
+          python-version: "3.12"
+
+      - name: Run harness + gate per skill
+        id: scan
+        if: steps.list.outputs.skills != '[]'
+        env:
+          SKILLS: ${{ steps.list.outputs.skills }}
+          HEAD_SHA: ${{ steps.pr.outputs.head_sha }}
+          BASE_SHA: ${{ steps.pr.outputs.base_sha }}
+          SKILL_BASE_DIR: ${{ github.workspace }}/skills
```
krollins-mdb, line 281: The companion harness therefore loads public evals.json files.

However, the companion task modules select hard-coded private case IDs:

- Natural-language querying: 37–48
- Query optimizer: 37–38
- Search and AI: 37

The public files contain:

- Natural-language querying: 1–36
- Query optimizer: 1–8 and 20
- Search and AI: 1–15

Essentially, numeric case IDs are being used as both identity and selection policy across the two repositories and we need to rework this a bit.

Some things we could do:

- Make every workflow consume the same checkout-backed case source.
- Select functional cases using explicit metadata such as the presence of functional_check, a tag, or a named suite.
- Do not hard-code repository-specific numeric IDs in task modules.

.github/workflows/qa-pr-screen.yml
```diff
@@ -0,0 +1,180 @@
...
+#     Missing model credentials fail loud there (harness.py preflight), not here.
+
+on:
+  workflow_dispatch:
+    inputs:
+      skill:
+        description: "Skill to screen"
+        required: true
+        type: string
+      pr_number:
```
krollins-mdb, line 53: `pr_number` controls only where the result is posted. The evaluated code is independently selected from `qa-pr-screen.yml`.

Nothing verifies that the dispatch ref is the supplied PR’s head SHA. An operator can therefore evaluate main or another branch and post the result to a PR as though that PR had been screened.

When `pr_number` is provided we could:

- Resolve the PR’s current head SHA.
- Check out that exact SHA.
- Include the evaluated SHA in the report.
- Refuse to post if the evaluated SHA does not match the PR head.

A dispatch without `pr_number` can continue to evaluate an arbitrary trusted ref and print the report without commenting.

.github/workflows/qa-mutation-battery.yml
```diff
@@ -0,0 +1,239 @@
...
+      epochs:
+        description: >-
+          Repeats per sample. Default 5: below that, run-to-run noise sits above the
+          battery's detection threshold and the report is suppressed as unmeasurable —
+          a lower-k run spends the model calls and learns nothing.
+        required: false
+        default: "5"
+        type: string
+
+# One battery per skill at a time. Two concurrent batteries interleave their runs in the
```
krollins-mdb: This states that batteries are serialized per skill, but its concurrency groups are:

```
qa-mutation-battery-all
qa-mutation-battery-<skill>
```

An all-skills run and a targeted run can therefore evaluate and publish the same skill concurrently. Both may update the same baseline file in the private repository.

The publication retry in `qa-mutation-battery.yml`is also brittle. If git pull --rebase produces a conflict, Git remains in an active rebase. Subsequent iterations retry without aborting or resolving that state, so they cannot recover.

We should:

- Serialize mutation publication globally, or introduce a lock keyed by each expanded matrix skill.
- If retaining retries, distinguish push races from rebase conflicts.
- Abort a failed rebase before another retry.
- Avoid retrying conflicts automatically when both runs changed the same baseline.

krollins-mdb (changes requested):
Thank you for all of the work on this, @cbullinger! I have some suggested fixes for things. [This one](https://github.com/mongodb/agent-skills/pull/55/changes#r3857502411) needs particular attention.

Generally, this looks great and I'm excited for it to get merged!

.github/workflows/qa-eval.yml
cbullinger: the thinking was that I didn't want this check blocking a PR on an infra error (i.e. something outside the author's control). but it probably makes more sense for any infra failings should be marked as failures.

.github/workflows/qa-eval-status.yml
cbullinger: what a miss. okay, updating to align the seeder's grep with the resolver's pattern

testing/evals.schema.json
cbullinger: good catch. i'm updating with dupe detection per file with a corresponding update on the private evals side.

.github/workflows/qa-eval.yml
cbullinger: def a code smell. for context, the hard-coded lists are generated by `scripts/convert-qa-cases.py`, which auto-detects the functional cases at generation time. i'm fixing so that the public `evals.json` becomes the single case source, and updating corresponding logic on the private evals side

.github/workflows/qa-mutation-battery.yml
cbullinger: updated per your rec

.github/workflows/qa-pr-screen.yml
cbullinger: updated

krollins-mdb (approved):
LGTM! Thank you, @cbullinger! 🎉

PR #56, opened 2026-08-02
chore: add plugin readmes

nirinchev:
Adds readmes for the two plugins - those are fairly barebones, but I guess it's better than nothing.


dylan-good-e (commented):
LGTM

dylan-good-e (approved):
LGTM

PR #62, opened 2026-08-28
feat(qa): publish actionable eval diagnostics

cbullinger:
## Summary

- publish sanitized per-case QA evaluation diagnostics instead of verdict-only artifacts
- isolate each skill run to its exact Inspect log and validate reports before upload
- add representative ASP trajectory checks and allow required Atlas project discovery

## Security

Raw Inspect logs are not uploaded. Reports pass through the private harness public-field allowlist and secret scrubber before reaching this public repository.


cbullinger:
Superseded by a clean branch based on current main.

PR #63, opened 2026-08-28
feat(qa): publish actionable eval diagnostics

cbullinger:
## Summary

- publish sanitized per-case QA evaluation diagnostics instead of verdict-only artifacts
- isolate each skill run to its exact Inspect log and validate reports before upload
- add representative ASP trajectory checks and allow required Atlas project discovery

## Security

Raw Inspect logs are not uploaded. Reports pass through the private harness public-field allowlist and secret scrubber before reaching this public repository.


.github/workflows/qa-eval.yml
```diff
@@ -314,24 +335,84 @@ jobs:
...
-              --arg runAt="$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
-              '{skillName: $skill, headSha: $headSha, baseSha: $baseSha, model: $model,
+              --arg runAt "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
+              '. + {skillName: $skill, headSha: $headSha, baseSha: $baseSha, model: $model,
                 verdict: $verdict, runAt: $runAt}' \
-              > "$RUNNER_TEMP/reports/qa-eval-report-$SKILL.json"
+              "$DETAIL_REPORT" > "$REPORT_TMP"; then
+              REPORT_VALID=false
+            fi
+            if [ "$REPORT_VALID" = true ] && ! jq -e \
```
krollins-mdb: The report validation is somewhat shallow. A sanitizer regression in the private harness could produce an unsafe artifact that this check accepts.

Validation restricts sample field names but does not validate their types, nested keys, or contents. Fields like trajectory, trajectoryChecks, judgeSummary, and validity can contain arbitrary strings or objects—including transcript fragments, tool arguments, paths, or secrets—and still pass validation before being uploaded publicly.

Recommendation:

1. Define the exact public report schema.
2. Validate each allowed field’s type and, for objects, its permitted nested keys.
3. Remove fields that are not explicitly intended for public artifacts.
4. Add negative tests to prevent future regressions.

krollins-mdb (approved):
LGTM! Just one edge case thing.

.github/workflows/qa-eval.yml
cbullinger: agreed. reworked it to emit only an explicit typed public schema

